PR #331, opened 2020-12-08
Plookup implementation

ndbunner:
Still working on some things with the prover and verifier, but would like some help getting the table and such from the composer. @Bounce23 do you have time for a chat this week?


Bounce23:
@ndbunner, yes I do. I'll contact you by other means asap.

PR #338, opened 2021-01-05
Refactor composer for Plookup

### LINKED ISSUE #326: Implement Lookup gate
The Lookup function will require a gate in the composer. 

The functionality will be something as follows: 
- Ensure lookup table is created
- Add selector polynomial
- Store table index inside 4th wire
- Constrain the gate equation to omit addition and multiplication selectors

This lives inside epic #325.

Bounce23:
The composer has been refactored to accept the plookup gate functionality.

This closes #326 

This PR will include the changes to the mathematical proof structure, so that PLONK provers accept PLOOKup arguments.


Bounce23:
The proof system is complete but there are some prupose full mistakes, all to do with h & t. Not big ones - I created and filled in the functions according to the new proof arguments, so it still makes sense, as in the expected values are taken into each fucntion. It's just about where these values can come from for the verifier. 
For instance, we the verifier uses h1 as an affine point in G1 for the lineariser commitment. As the prover doesn’t make a commitment to h1 in the proof, where should this come from? Do we recompute f & t and then s, and then half? Else should h1 be in the proof commitments? 
Another one is when the verifier is creating these tables they can’t come from the composer, perhaps we need the verify to add in a ‘preprocessedtable’ like what i did in preprocess.rs to get comms and evals for t_1.

Bounce23:
This PR can be closed as all of the changes caused errors in the fundamentals of the back end. 
All of these changes are included, without errors, in PR #387.

PR #347, opened 2021-01-13
Include plookup in opening proof polys

lopeetall:


lopeetall:
Changes I made here affect many tests which are now failing. Will work on updating the old tests next.

Bounce23:
I would say until this is ready to go into `dusk-network:composer_plookup`, then we should turn it to draft.

lopeetall:
It looks like the tests that fail are those that involve proof/proverkey serialization or those that run a verifier. I can push some new changes that fix the serialization issues if that is helpful.

Bounce23:
Closed this PR as #366 was merged containing these changes.

PR #349, opened 2021-01-13
Heavy data structures from unchecked

vlopes11:
#332

### DIFF
--- a/src/serialisation.rs
+++ b/src/serialisation.rs
@@ -54,7 +54,7 @@ pub fn read_g1_affine(bytes: &[u8]) -> Result<(G1Affine, &[u8]), Error> {
     let (bytes48, rest) = read_n(48, bytes)?;
     let mut arr48 = [0u8; 48];
     arr48.copy_from_slice(bytes48);
-    let g1 = G1Affine::from_compressed(&arr48);
+    let g1 = G1Affine::from_bytes(&arr48);
     if g1.is_none().into() {
         return Err(SerialisationErrors::PointMalformed.into());
     }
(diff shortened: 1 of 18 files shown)


src/serialisation.rs
```diff
@@ -34,15 +35,13 @@ pub fn read_n(n: usize, bytes: &[u8]) -> Result<(&[u8], &[u8]), Error> {
...
-    let mut arr32 = [0u8; 32];
-    arr32.copy_from_slice(bytes32);
-    let scalar = BlsScalar::from_bytes(&arr32);
-    if scalar.is_none().into() {
-        return Err(SerialisationErrors::BlsScalarMalformed.into());
-    }
-    Ok((scalar.unwrap(), rest))
+    let (bytes, rest) = read_n(BlsScalar::SIZE, bytes)?;
+
+    BlsScalar::from_slice(bytes)
```
ZER0, line 40: I'm okay to keep as is for the time being but this is not optimal, since we're checking for the length twice, in `read_n` and in `from_slice`.
Probably the best approach would be remove `read_n` from the `read_*` where we're using `from_slice` and just mapped `BadLength` to `NotEnoughBytes` and `InvalidData` to the current type error (e.g. `SerialisationErrors::BlsScalarMalformed`) (if we don't want to change too much the API, I mean).

ZER0 (approved):
Looks good to me. I think we should redone the serialisation module; but it's not something we should do right now.
The only thing that bother me is the double check on `read_n` and `from_slice` (see my comment), but we can probably live with that.

src/serialisation.rs
vlopes11: I agree with your concern, and this specific use-case is interesting to add as a feature to `dusk-bytes`.

In this usage, we need to not only parse the targeted type, but also split the slice into the tuple composed by the used bytes and the remainder ones.

This pattern is broadly used in crates such as `nom` and will be very helpful to us!

vlopes11: https://github.com/dusk-network/dusk-bytes/issues/11

PR #357, opened 2021-01-18
Replace omega_roots with scalar 1

lopeetall:
In the plookup paper and our document with code changes a factor of `x-omega^n` (or something similar) is included in several places where computing plookup permutation checks. Since `omega` is an nth root of unity, `omega^n` equals 1. Here `omega^n` has been replaced with `BlsScalar::one()` and references to `omega_roots` have been removed.


Bounce23:
This catch is good, the wrapping back around to one should always be the case without the need to call the last domain element.

Bounce23:
We can discuss tonight on the call to merge this in and then rebase some things when we merge the other PR's and then fix tests.

Bounce23:
Will close this PR as #366 was merged.

PR #387, opened 2021-02-04
Modularise plookup

### LINKED ISSUE #379: Separate StandardComposer from shared functionality on the base operations
Implement a new base composer which encompasses all the operations for Standard and Plookup Composers.

Bounce23:
This PR abstracts over two composers to make plonk with looks up forward compatible for all future specialised composers. 
This closes #379.


Bounce23:
This PR was created to incorporate the contents of #338 and #330 without the mass issues to the API, proving system or back end. Now, all of the tests pass for PLONK and PLONK with lookups (Plookup).

Bounce23:
The plookups only pass if you hard code the polynomials, else the commit keys return them a `poly of degree 0`. This is due to the inability of KZG10 with PLONK to take in arbitrary polynomials. 

The plookups also only pass if you hardcode the padding of the tables. So they are padded, as extra wires to n. Where n is the circuit size. If you fix the entries in the queries table to be greater than the entries in the lookup table, it passes. 
The reason for this is the prover expects now all 3 different representations of wires to be of equal length. 

This is being worked on by me and @lopeetall

Bounce23:
The error behind the zero polys has been solved. This is for both prover and verifier. 
Additionally, the composer has been optimised by removing query wires and selecting them from lookup_selector valid gates with 
```let f_1_scalar = w_l_scalar.iter().zip(&self.cs.q_lookup).map(|(w, s)| w*s + (BlsScalar::one()-s) * lookup_table.t_1.0.0[0]).collect::<Vec<BlsScalar>>();```

Bounce23:
In line with #379, this PR is to be closed.

PR #395, opened 2021-02-10
Modularise plookup

lopeetall:


Bounce23:
The failure here is just from a header, I've fixed it at the base branch. If you pull @lopeetall then analyser should work.

Bounce23 (approved):
The compression of tables within the proving system should be done with a `pub fn` from the `multiset.rs`

However, for the purpose of debugging, having it as an open function is fine but the changes should be made before a merge into master.

PR #410, opened 2021-02-16
Modularise plookup

lopeetall:


src/constraint_system/plookup_composer.rs
```diff
@@ -469,7 +469,7 @@ mod tests {
         let mut prover = PlookupProver::new(b"demo");
```
Bounce23, line 469: Looks good to me, could be worthwhile marking this is a test circuit. So that we can add all the different types of circuits into integration tests, for benchmarking later. If not, it can be done later - up to you.

Bounce23 (approved):
As we worked on this offline together I have only one nit, feel free to address - if not, it can be done on integration tests branch.

src/constraint_system/plookup_composer.rs
Bounce23: After some discussion with @lopeetall, we have decided to merge and port this issue to integration tests.

PR #422, opened 2021-03-02
Add table to composer

lopeetall:


Bounce23 (approved):
This looks good to me @lopeetall, feel free to merge

If possible we could create/assign issues in our next meeting then we have a cohesive direction for the coding. I think this will be made easier with you having write access.

PR #425, opened 2021-03-03
Add plookup tests

lopeetall:
`test_plookup_all_gates` combines plookup gates with all base gates from other widgets (arithmetic, range, logic, boolean). Ecc gates are not included since these simply resolve to arithmetic gates.

### DIFF
--- a/src/constraint_system/boolean.rs
+++ b/src/constraint_system/boolean.rs
@@ -4,8 +4,8 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-use crate::constraint_system::{PlookupComposer, StandardComposer};
 use crate::constraint_system::Variable;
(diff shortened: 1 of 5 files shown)


lopeetall:
> `test_plookup_all_gates` combines plookup gates with all base gates from other widgets (arithmetic, range, logic, boolean). Ecc gates are not included since these simply resolve to arithmetic gates.

Unfortunately this statement wasn't correct. There is a `fast_add` gate and `big_mul_gate` which do not work with plookup currently. I will adjust them so they work with plookup and add these gates into the tester as well so they are not overlooked.

Bounce23:
@lopeetall, can this PR be closed or should it be caught up with the latest branch of `modularise_plookup` and amended for the final tests?

Bounce23 (commented):
PR is to be closed as the tests were fixed elsewhere.

PR #530, opened 2021-05-07
Refactor Hash constraint tables

### LINKED ISSUE #527: Implement Fixed lookup tables for reinforced concrete
The current implementation progress of Zelbet hash function requires fixed lookup tables to provide low degree non-linear polynomial evaluations. 

These are extracted from 3 different Lookup tables, each for the gate entries. 
These ought to be made as part of this hash function operation.

### LINKED ISSUE #528: Construct T1 table for reinforced concrete
The reinforced concrete paper requires particular lookup tables. The first is a functioned binary table. This needs to be implemented in the Lookup section and should be made public.

### LINKED ISSUE #529: Construct T2 table for reinforced concrete
The reinforced concrete paper requires particular lookup tables. The second table is a iterative function table, which takes a dynamic input. This needs to be implemented in the Lookup section and should be made public.

Bounce23:
As per the reinforced concrete tables, the new tables will be used to ensure
certain gate checks are made. This is now done for three separate hashing-tables.

This closes #528 
This closes #529


src/constraint_system/composer.rs
```diff
@@ -574,13 +574,21 @@ impl StandardComposer {
 
         c
     }
+
+    /// When StandardComposer is initialised, it spawns a dummy table
+    /// with 3 entries that should not be removed. This function appends
+    /// its input table to the composer's dummy table
+    pub fn append_lookup_table(&mut self, table: &PlookupTable4Arity) {
+        (0..table.0.len()).for_each(|k| self.lookup_table.0.push(table.0[k]))
```
julesdesmit, line 582: Refactor this to iterate over the table directly:

```rust
table.0.iter().for_each(|k| self.lookup_table.0.push(k))
```

It's a bit cleaner :)

src/constraint_system/composer.rs
```diff
@@ -610,26 +618,72 @@ mod tests {
...
+    #[should_panic]
+    fn test_gadget_fail() {
+        let mut t = PlookupTable4Arity::new();
+        t.insert_special_row(
+            BlsScalar::from(12),
+            BlsScalar::from(12),
+            BlsScalar::from(12),
+            BlsScalar::from(12),
+        );
+        let res = gadget_plookup_tester(
```
julesdesmit, line 676: Should this result not be checked?

src/constraint_system/composer.rs
```diff
@@ -736,7 +791,8 @@ mod tests {
     }
 
     #[test]
-    // XXX: Move this to integration tests
+    #[ignore]
+    // XXX: Test works in principle, fails due to table size Vs number of lookups
```
julesdesmit, line 795: Make sure this is extracted into a separate issue so it doesn't get lost.

src/constraint_system/zelbet.rs
```diff
@@ -8,151 +8,266 @@
...
+        assert!(res.is_ok());
+    }
+
+    #[test]
+    fn test_s_box_and_constraints_fails() {
+        let res = gadget_tester(
+            |composer| {
+                let hash_table = PlookupTable4Arity::create_hash_table();
+                composer.append_lookup_table(&hash_table);
+                let one = composer.add_input(BlsScalar::from(100));
```
julesdesmit, line 204: Bit of a nitpick :) but shouldn't this be one_hundred?

src/proof_system/prover.rs
```diff
@@ -156,7 +156,12 @@ impl Prover {
 
         // Sort table so we can be sure to choose an element that is not the highest or lowest
         compressed_t.sort();
-        let second_element = compressed_t[1];
+        let mut second_element = compressed_t[1];
+        let mut k = 1;
+        if second_element == compressed_t[0] {
+            k = k + 1;
+            second_element = compressed_t[k];
+        }
```
julesdesmit, line 164: You can make this more concise as well:

```rust
let second_element = match compressed_t[0] == compressed_t[1] {
    true => compressed_t[2],
    false => compressed_t[1],
}
```

julesdesmit (changes requested):

src/constraint_system/composer.rs
Bounce23: This is added with a `#[should_panic]` and assert!(res.is_err());
As the element is missing from the table

src/constraint_system/composer.rs
Bounce23: okay

src/proof_system/prover.rs
```diff
@@ -158,10 +158,11 @@ impl Prover {
         compressed_t.sort();
         let mut second_element = compressed_t[1];
         let mut k = 1;
```
julesdesmit, line 160: You can remove lines 159 and 160.

julesdesmit (approved):
One more thing. When that is addressed, fine to merge!

PR #532, opened 2021-05-20
Test hash

Bounce23:
This PR implements the exposing functionality on Zelbet, from PLONK. 

This is done by adding composer gadgets which expose the inner scalar, for modular reduction in circuit.


src/constraint_system/composer.rs
```diff
@@ -417,6 +417,112 @@ impl StandardComposer {
...
+    /// is satisfied for each one of the [`StandardComposer`]'s gates.
+    ///
+    /// The recommended usage is to derive the std output and the std error to a
+    /// text file and analyze there the gates.
+    ///
+    /// # Panic
+    /// The function by itself will print each circuit gate info until one of
+    /// the gates does not satisfy the equation or there are no more gates. If
+    /// the cause is an unsatisfied gate equation, the function will panic.
+    pub fn check_circuit_satisfied(&self) {
```
julesdesmit, line 431: This was already in PLONK, so I'm not sure how it happened that this suddenly got added as new. I guess because the branches diverged before this feature was included? If possible, it may be better to leave this out, since you will get a conflict out of it when trying to rebase on master, and you'll probably inadvertently remove the original author from these lines.

src/constraint_system/zelbet.rs
```diff
@@ -0,0 +1,1017 @@
...
+                );
+            },
+            800,
+        );
+        assert!(res.is_ok());
+    }
+}
+
+/// decomposition = [s_n, s_{n-1} ..., s_1]
+pub const DECOMPOSITION_S_I: [BlsScalar; 27] = [
```
julesdesmit, line 159: Move these constants into their own file, just a bit easier to get around that way :)

julesdesmit (changes requested):

src/constraint_system/composer.rs
Bounce23: It's better to 'allow incoming' when this goes into master. 
They were moved when the features wouldn't work with the new proof system, by another developer.
I'll make a note of allowing master to overwrite these lines.

julesdesmit (approved):
LGTM, merge when CI passes.

PR #543, opened 2021-06-23
Fix check_circuit_satisfied

### LINKED ISSUE #541: Missing qfixed and qvar in check_circuit_satisfied
### What's wrong

The variable k in the check_circuit_satisfied should check if constraints related to q_fixed_group_add and q_variable_group_add selectors are satisfied, but the check is unimplemented now.

### How can we fix it

Add qfixed and qvar checks to the variable k

ChihChengLiang:
### What's wrong

Fix #541

### How we are fixing this

- [x] create a test to make sure it fails when one of the `qfixed` and `qvar` are switched one
- [ ] fix qfixed
- [ ] fix qvar

### DIFF
--- a/src/constraint_system/composer.rs
+++ b/src/constraint_system/composer.rs
@@ -563,9 +563,12 @@ impl StandardComposer {
             - q_variable_group_add -> {:?}\n
             # Witness polynomials:\n
             - w_l -> {:?}\n
+            - w_l_next -> {:?}\n
             - w_r -> {:?}\n
+            - w_r_next -> {:?}\n
             - w_o -> {:?}\n
-            - w_4 -> {:?}\n",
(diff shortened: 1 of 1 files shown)


src/constraint_system/composer.rs
```diff
@@ -492,7 +492,7 @@ impl StandardComposer {
     /// The function by itself will print each circuit gate info until one of
     /// the gates does not satisfy the equation or there are no more gates. If
     /// the cause is an unsatisfied gate equation, the function will panic.
-    #[cfg(feature = "trace")]
+    // #[cfg(feature = "trace")]
```
ChihChengLiang, line 495: disable the trace for the test to work for now

moCello:
Closing as not applicable anymore

PR #546, opened 2021-07-06
Fix query poly

lopeetall:
The previous system used two slightly different versions of the query polynomial (`f` in the plookup paper) at different places in the proving system to solve a discrepancy found in the original paper. These were called `f_short` and `f_long`. However there was nothing to enforce that these polynomials were actually related.

Since `f_short` is essentially a one-element shift of `f_long`, except for one difference in the first element, the previous checks that enforce relationships could be easily modified. The modified checks now use `f` and `f_next` in place of `f_long` and `f_short` respectively. Then the `f` poly was added to the shifted aggregate witness so that `f` is checked against `f_next` in the same way other shifted polynomials are checked.

### DIFF
--- a/src/proof_system/linearisation_poly.rs
+++ b/src/proof_system/linearisation_poly.rs
@@ -69,10 +69,10 @@ pub struct ProofEvaluations {
     pub h_2_next_eval: BlsScalar,
 
     /// Evaluations of the query polynomial at `z`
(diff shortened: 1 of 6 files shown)


lopeetall:
This is obviated by https://github.com/dusk-network/plonk/pull/547 and is no longer needed

PR #547, opened 2021-07-14
Make alternating

lopeetall:
The plookup permutation product argument is changed to use an alternating split for h1 and h2, and the prover and verifier algorithms have been modified to match.


Bounce23 (approved):
This PR looks good to me, so feel free to merge whenever. 

All of the comment changes as well as the actual polynomial arguments themselves seem to be well changed inline with the new paper.

PR #550, opened 2021-07-26
Multiplications replaced with hard coded results

EDGDrummond:
Modular reductions leave a maximum value of 703,
so rather than convert this result to Montgomery
form with a multiplication it can be hard coded.


Bounce23 (changes requested):
Apply a rust or cargo fmt to this PR and if this passes, request a review @EDGDrummond

Bounce23 (approved):
LGTM! 
Feel free to merge

PR #551, opened 2021-07-28
Add review suggestions

lopeetall:
This code implements the new alternating method for constructing `h_1` and `h_2` in the plookup permutation argument, two reviewer suggestions, and reenables lookup table preprocessing.


Bounce23 (approved):
Having already gone through these changes, the implementation looks good to me so feel free to merge.

PR #591, opened 2021-09-27
Add codecoverage to `add_lookups`

ureeves:


ureeves:
This seems to generate good results. Currently we're at about ~75%. Should be enough to view which changes were introduced that are untested in the target branch.

vlopes11 (dismissed):
LGTM. Nit: fix broken line breaks at the end of the file

ureeves:
Reformed the CI a little bit by having two matrices - one for building for `no_std` environments, and another for testing in `std` (--all-features) environments.

.github/workflows/dusk_ci.yml
```diff
@@ -19,73 +18,108 @@ jobs:
...
+      - name: Add target
+        run: rustup target add ${{ matrix.target }}
+
+      - name: Build project with alloc
+        uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --release --no-default-features --features alloc --target thumbv6m-none-eabi
-      - uses: actions-rs/cargo@v1
+          args: --release --no-default-features --features alloc --target ${{ matrix.target }}
```
vlopes11, line 70: If we are just building without any tests, then `release` is just additional overhead that brings no benefit

.github/workflows/dusk_ci.yml
```diff
@@ -19,73 +18,108 @@ jobs:
...
+
+      - name: Build project without alloc
+        uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --release --no-default-features --target thumbv6m-none-eabi
+          args: --release --no-default-features --target ${{ matrix.target }}
```
vlopes11, line 76: Ditto; `release` without tests

.github/workflows/dusk_ci.yml
```diff
@@ -19,73 +18,108 @@ jobs:
...
+        with:
+          command: test
+          args: --release --all-features
+
+      - name: Test project generating coverage data
+        if: ${{ matrix.os == 'ubuntu-latest' && matrix.toolchain == 'nightly' }}
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --release --all-features --no-fail-fast
```
vlopes11, line 111: Do we keep the proper debug symbols in release mode? Just curiosity; I think the coverage tool requires precise debug data in order to evaluate the source mappings

vlopes11 (commented):
Some questions

.github/workflows/dusk_ci.yml
ureeves: In this case I just kept `release` for the sake of keeping it like it was before. A pro is that it is more consistent with all else that is happening, but I agree right now it adds no tangible benefit.

.github/workflows/dusk_ci.yml
ureeves: Same as above

.github/workflows/dusk_ci.yml
ureeves: I force building with debug symbols in the flags below. Specifically `-Cdebuginfo=2`. That way we keep the test speed benefit of `release` while also being able to profile the code.

.github/workflows/dusk_ci.yml
ureeves: One thing that I would say as well is that the building incurs much less of a cost than running the tests, and since the jobs run in parallel this is not noticeable.

PR #592, opened 2021-09-28
Add missing dummy constraints in test

ureeves:
Adds missing dummy constraints in test and removes the `#[ignore]` resulting in a passing test.

### DIFF
--- a/src/constraint_system/composer.rs
+++ b/src/constraint_system/composer.rs
@@ -827,12 +827,12 @@ mod tests {
 
     #[allow(unused_variables)]
     #[test]
-    #[ignore]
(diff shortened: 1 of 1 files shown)


src/constraint_system/composer.rs
```diff
@@ -827,12 +827,12 @@ mod tests {
 
     #[allow(unused_variables)]
     #[test]
-    #[ignore]
     /// Tests that an empty circuit proof passes
-    fn test_prove_verify() {
+    fn test_empty_circuit() {
```
vlopes11, line 831: Its not really empty but with dummy constraints. I believe the name should reflect that

Issue #590, opened 2021-09-27
Plonkup - Fix or document ignored tests


vlopes11:
**Describe the bug**
Some of the tests are ignored without a comment explaining why. We should either fix the problems related to these test cases or explicitly document why they should be ignored

**To Reproduce**
N/A

**Expected behaviour**
Every test case should be passing

**Logs/Screenshot**
N/A

**Platform**
N/A

**Additional context**
N/A

ureeves:
Closed by #593

PR #593, opened 2021-09-29
Fix ignored tests

ureeves:
Addresses issue #590

`test_plonkup_proof` was pretty interesting. Firstly `max_degree` and `truncated_degree` were set far too small, secondly the circuits definitions were set differently for the `Prover` and for the `Verifier`.


vlopes11 (approved):
LGTM

PR #601, opened 2021-10-01
Add public unsafe `evaluate_witness()` to the composer

ZER0:
- Add public `duplicate_witness()` to the composer
- Fix benches
- Fix few nits


src/constraint_system/composer.rs
```diff
@@ -128,13 +128,23 @@ impl TurboComposer {
...
+    /// # Safety
+    ///
+    /// Using the allocated scalar of a witness inside a gadget could produce
+    /// an unsound circuit.
+    pub unsafe fn evaluate_witness(&self, witness: &Witness) -> &BlsScalar {
+        &self.witnesses[witness]
+    }
+
+    /// Add a new witness with the same scalar of the given one
+    pub fn duplicate_witness(&mut self, witness: &Witness) -> Witness {
```
vlopes11, line 141: Please, elaborate on why this functionality is needed.

This is not particularly secure because the new witness isn't constrained to the previous one. We often want to avoid at all costs creating new witnesses, and we can do so only if we have a clear constraint that will make that operation safe.

src/circuit.rs
```diff
@@ -301,6 +301,7 @@ where
     }
 
     /// Return the list of public inputs generated by the gadget
+    #[allow(clippy::wrong_self_convention)]
```
vlopes11, line 304: Oh, my bad there. Maybe rename to `to_public_inputs`?

src/constraint_system/composer.rs
```diff
@@ -128,13 +128,23 @@ impl TurboComposer {
...
 
     /// Evaluate the runtime value of a witness
-    #[allow(dead_code)]
-    pub(crate) fn evaluate_witness(&self, witness: &Witness) -> &BlsScalar {
-        &self.witnesses[&witness]
+    ///
+    /// # Safety
+    ///
+    /// Using the allocated scalar of a witness inside a gadget could produce
+    /// an unsound circuit.
```
vlopes11, line 135: I think this could be reformulated.

The problem is that scalar evaluation on gadget construction isn't expected and could lead to different circuit representations for the same code.

If the user calls this function, he should be aware it is to perform operations outside the circuit that have no security value at all.

src/fft/domain.rs
```diff
@@ -406,7 +406,7 @@ mod tests {
 
     #[test]
     fn dusk_bytes_evaluation_domain_serde() {
-        let eval_domain = EvaluationDomain::new(1 << 13 - 1)
+        let eval_domain = EvaluationDomain::new((1 << 13) - 1)
```
vlopes11, line 409: Your change improves the code, but seems it should be `1 << 13`. We expect here a power of 2 and this `- 1` simply doesn't make any sense.

src/fft/polynomial.rs
```diff
@@ -109,7 +109,7 @@ impl Polynomial {
             .map(|(c, p)| p * c)
             .collect();
         let mut sum = BlsScalar::zero();
-        for eval in p_evals.into_iter() {
+        for eval in p_evals {
```
vlopes11, line 112: Maybe we also change to Rust 2021?

vlopes11 (changes requested):
Few nits, but looks good overall

src/constraint_system/composer.rs
ZER0: In zelbet we need to append a new witness based on the value of another witness. One solution is indeed having:
```rust
let w1 = composer.append_witness(unsafe { composer.evaluate_witness(w) });
```

But since we don't actually care about leaking the allocated scalar – we do not perform operation with that, we just care about the witness – having a function that avoid to expose the scalar sounded better.

but we don't actually care what is the allocated scalar.

src/fft/domain.rs
ZER0: We have two issues here then.
1. We should ensure a failure if we do not give a power of two: but the tests are passing.
2. Before it was basically `1 << 12` and the tests were still passing. So either we need `1 << 12` and not `1 << 13`, or our code and tests are wrong.

src/fft/polynomial.rs
ZER0: We should change to the `2021` edition once it its stable, to be on the safe side. I believe it's planned by the end of the month.

In addition, there are still few things in plonkup that are not following the `2018` edition, so we should ensure that first.

src/constraint_system/composer.rs
ZER0: `scalar evaluation on gadget construction isn't expected and could lead to different circuit representations for the same code` to me is having a circuit that is unsound (in the rust terminology). I'm happy to reformulate / adding more information. What about:

```
Witness evaluation inside a gadget isn't expected and could produce an unsound circuit (different circuit representation for the same code).

Calling this function performs operations outside the circuit.
```

If you want to add the part `that have no security value at all` I think we should expand the concept.

Personally I don't think we should mention how to prove and verify here,  because if get to read this comment but you don't know even how to prove and verify in plonk it means you just skipped all the rest of the library / tests.

src/circuit.rs
ZER0: This is a draft PR, I was planning to go over all the `allowing` part – at least for the things introduce in the branch I'm merging  – since there are other issue like that around the code base, and we could have an issue to tackle them all together.

src/constraint_system/composer.rs
vlopes11: Since `duplicate_witness` is also unsafe (e.g. the previous constraints are not applied to the new witness and the circuit will still be valid), the snippet suggestion is a better approach

src/constraint_system/composer.rs
vlopes11: That sounds good

src/fft/domain.rs
ZER0: For the time being I'm going to use directly `1 << 12` as discussed, however we would need to improve the API that takes numbers that needs to be power of two / next power of two.

src/constraint_system/composer.rs
```diff
@@ -994,7 +1001,8 @@ mod tests {
         // So pre-fetch these before calling Prove
         let public_inputs = prover.cs.into_dense_public_inputs();
 
-        (prover.prove(&ck).unwrap(), public_inputs);
+        prover.prove(&ck).unwrap();
+        drop(public_inputs);
```
ureeves, line 1005: `public_inputs` seems to already drop at the end of this scope. Why drop it explicitly?

ureeves (commented):
Just a nit on an explicit drop. Maybe it's necessary.

Issue #602, opened 2021-10-06
Ensure our code coverage provides accurate results


ZER0:
**Describe what you want implemented** 
Currently we're not sure if the code coverage numbers are accurate enough (however, they should be on the pessimistic side)

**Describe "Why" this is needed**
The library needs to be properly tested, so we need to know the actual code coverage.

**Describe alternatives you've considered**
No code coverage.

**Additional context**
N/A

PR #604, opened 2021-10-06
Improve codecov reports

ureeves:
This is an effort to ensure the coverage results are correct. As the issue #602 states we had a suspicion that the CI was producing bad results - specifically on the pessimist side.

As this PR shows, using `kcov` - the tool that was previously used to generate reports when Travis CI was still in use - shows an incredible bump in results.


ureeves:
These results honestly make much more sense to me, but need review.

Issue #589, opened 2021-09-27
Plonkup - Clarify the arity structure for the tables


vlopes11:
**Describe what kind of specification you want to have created**
Currently `plonkup` added `PlonkupTable3Arity` and `PlonkupTable4Arity`. However, the turbo composer is implemented to support only `PlonkupTable4Arity` as in https://github.com/dusk-network/plonk/blob/add_lookups/src/constraint_system/composer.rs#L102

We need:

1) Understand if we have a clear benefit or use case to have multiple arity definitions
2) Generalize the test cases in https://github.com/dusk-network/plonk/blob/add_lookups/src/plonkup/table/lookup_table.rs#L585 to cover every table implementation (currently its covering only `PlonkupTable3Arity`

marta-belles:
Current documentation of `plonkup` contains a section explaining how to handle multiple tables which does not match the current implementation. The document should also be changed according to the code.

PR #619, opened 2021-10-14
Remove `Plonkup3Arity`

ureeves:
As discussed this remove the `Plonkup3Arity` and makes the code use and test only `Plonkup4Arity` now renamed to `LookupTable`.

* Remove `PlonkupTable3Arity`
* Change to use `PlonkupTable4Arity`
* Rename `PlonkupTable4Arity` to `LookupTable`
* Fix tests to work with `LookupTable`
* Remove bigint crate and the usage of `u256` from plonkup 

See also: #589

### DIFF
--- a/src/plonkup.rs
+++ b/src/plonkup.rs
@@ -17,6 +17,6 @@ pub(crate) mod table;
 pub use multiset::MultiSet;
 pub use table::hash_tables::constants;
 pub use table::{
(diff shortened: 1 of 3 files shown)


ureeves:
> Looks good overall but the name `IndexTable` isn't descriptive enough. Any particular reason this isn't named `LookupTable`?

That is a better name 🤦‍♂️. Fixed

PR #625, opened 2021-10-15
Change to use `From<JubJubScalar>` for BlsScalar

ureeves:
Change all conversions to `BlsScalar` from `JubJubScalar` to use the `From`
implementation.

Resolves: #294

### DIFF
--- a/src/constraint_system/ecc/scalar_mul/fixed_base.rs
+++ b/src/constraint_system/ecc/scalar_mul/fixed_base.rs
@@ -175,8 +175,7 @@ mod tests {
                     0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
                     0, 0,
                 ]);
-                let bls_scalar =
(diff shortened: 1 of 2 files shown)


vlopes11:
> Can you think of a situation where they might go wrong?

Nothing can really go wrong there because we are using affine points so worst case scenario we have points that aren't in the curve and this would lead to broken circuits (and this is the expected behavior if the witnesses are invalid).

However, if it was extended, then we need to perform additional checks to prevent `panic`. The `from` API already performs all these checks for us and here we are basically skipping it. In this case, its just a matter of API consistency.

ureeves:
> > Can you think of a situation where they might go wrong?
> 
> Nothing can really go wrong there because we are using affine points so worst case scenario we have points that aren't in the curve and this would lead to broken circuits (and this is the expected behavior if the witnesses are invalid).
> 
> However, if it was extended, then we need to perform additional checks to prevent `panic`. The `from` API already performs all these checks for us and here we are basically skipping it. In this case, its just a matter of API consistency.

I suggest we merge and file an issue about this, since this PR is about conversions between scalars and not affine.

Issue #567, opened 2021-09-10
Documentation not rendering math


ureeves:
The equations and other math does not render in the "notes" part of the documentation. As an example take look at the current state of [docs.rs](https://docs.rs/dusk-plonk/0.8.1/dusk_plonk/notes/prove_verify/index.html).

moCello:
Describe what you want implemented
Include Katex-html header in the generated docs

Describe "Why" this is needed
This is needed in order to render LaTeX correctly in the official documentation of our crate

Describe alternatives you've considered
Leave as is which means that the math in our documentation is unreadable

Additional context
N/A

moCello:
Re-opening issue because LaTeX equations still don't render on docs.rs (they do locally). Inline LaTeX doesn't render both locally and on docs.rs.

PR #630, opened 2021-11-11
Add support for rendering LaTeX in the docs

moCello:
Previously LaTeX wasn't rendered in the official documentation of
our crate. With this change a custom header is inserted in the
html-files for the documentation, allowing LaTeX generated code
to be rendered with KaTeX.

As a side-effect of this change, `cargo doc` needs to be run with
the `--no-deps` flag:
```
$ cargo doc --no-deps
```

Resolves: #567


ureeves (approved):
LGTM

PR #651, opened 2021-12-14
Add blinding scalars

moCello:
Add blinding scalars in round 1, 2 and 3 of the proof


src/circuit.rs
```diff
@@ -256,7 +257,7 @@ where
         pub_params: &PublicParameters,
     ) -> Result<(ProverKey, VerifierData), Error> {
         // Setup PublicParams
-        let (ck, _) = pub_params.trim(self.padded_gates())?;
+        let (ck, _) = pub_params.trim(self.padded_gates() + 6)?;
```
moCello, line 260: Maybe add a comment why + 6

src/permutation.rs
```diff
@@ -212,14 +212,14 @@ impl Permutation {
...
-    pub(crate) fn compute_permutation_poly(
+    pub(crate) fn compute_permutation_vec(
         &self,
         domain: &EvaluationDomain,
         wires: [&[BlsScalar]; 4],
         beta: &BlsScalar,
         gamma: &BlsScalar,
         sigma_polys: [&Polynomial; 4],
-    ) -> Polynomial {
+    ) -> Vec<dusk_bls12_381::BlsScalar> {
```
moCello, line 222: You can use `Vec<BlsScalar>` and don't have to mention the `dusk_bls12_381` module path since it is imported on the top of the file already.

More generally though, why do you not use `Polynomial` here? We have to convert the evals to to their polynomial form at some point, so why do not do it here?

src/permutation.rs
```diff
@@ -308,7 +308,7 @@ impl Permutation {
         h_2: &[BlsScalar],
         delta: &BlsScalar,
         epsilon: &BlsScalar,
-    ) -> Polynomial {
+    ) -> Vec<dusk_bls12_381::BlsScalar> {
```
moCello, line 311: Same as above

src/proof_system/preprocess.rs
```diff
@@ -121,168 +121,168 @@ impl TurboComposer {
         let (_, selectors, preprocessed_table, domain) =
             self.preprocess_shared(commit_key, transcript)?;
 
-        let domain_4n = EvaluationDomain::new(4 * domain.size())?;
-        let q_m_eval_4n = Evaluations::from_vec_and_domain(
-            domain_4n.coset_fft(&selectors.q_m),
-            domain_4n,
+        let domain_8n = EvaluationDomain::new(8 * domain.size())?;
```
moCello, line 124: Maybe add a comment that explains the `8`

src/proof_system/prover.rs
```diff
@@ -145,6 +147,30 @@ impl Prover {
         self.preprocessed_transcript.append_message(label, message);
     }
 
+    /// Adds the blinding scalars to a given vector. Always the same elements
+    /// of 'w_vec' are modified at the beginning of it, and appended at the end:
+    /// if hiding degree = 1: (b2*X(n+1) + b1*X^n - b2*X - b1) + w_vec
+    /// if hiding degree = 2: (b3*X^(n+2) + b2*X(n+1) + b1*X^n - b3*X^2 - b2*X
+    /// - b1) + w_vec
+    pub(crate) fn blind_poly<R: RngCore + CryptoRng>(
+        w_vec: &Vec<dusk_bls12_381::BlsScalar>,
```
moCello, line 156: Above `dusk_bls12_381::BlsScalar` is already imported, so you can simply use `&Vec<BlsScalar>` here.

src/proof_system/prover.rs
```diff
@@ -145,6 +147,30 @@ impl Prover {
...
+        hiding_degree: usize,
+        domain: &EvaluationDomain,
+        mut rng: &mut R,
+    ) -> Polynomial {
+        let mut w_vec_i = domain.ifft(w_vec);
```
moCello, line 161: I find the names `w_vec` and `w_vec_i` not so descriptive. Isn't `w_vec_i` just `w_poly` and `w_vec` the poly in evaluation form? What about calling them `w_poly` and `w_poly_evals`?

src/proof_system/prover.rs
```diff
@@ -145,6 +147,30 @@ impl Prover {
         self.preprocessed_transcript.append_message(label, message);
     }
 
+    /// Adds the blinding scalars to a given vector. Always the same elements
+    /// of 'w_vec' are modified at the beginning of it, and appended at the end:
+    /// if hiding degree = 1: (b2*X(n+1) + b1*X^n - b2*X - b1) + w_vec
+    /// if hiding degree = 2: (b3*X^(n+2) + b2*X(n+1) + b1*X^n - b3*X^2 - b2*X
+    /// - b1) + w_vec
+    pub(crate) fn blind_poly<R: RngCore + CryptoRng>(
```
moCello, line 155: What is the reason you add the `blind-poly` to the `Prover`? Doesn't it make more sense to have it added to the `fft::Polynomial` type?

src/proof_system/widget/permutation/proverkey.rs
```diff
@@ -142,7 +142,7 @@ impl ProverKey {
             &self.s_sigma_4.0,
         );
 
-        let domain = EvaluationDomain::new(z_poly.degree()).unwrap();
+        let domain = EvaluationDomain::new(z_poly.degree() - 2).unwrap();
```
moCello, line 145: Add a comment on the ` - 2`

moCello (commented):
Left a few comments

src/circuit.rs
```diff
@@ -290,13 +291,14 @@ where
...
-    fn prove(
+    fn prove<R: RngCore + CryptoRng>(
         &mut self,
         pub_params: &PublicParameters,
         prover_key: &ProverKey,
         transcript_init: &'static [u8],
+        mut rng: &mut R,
     ) -> Result<Proof, Error> {
-        let (ck, _) = pub_params.trim(self.padded_gates())?;
+        let (ck, _) = pub_params.trim(self.padded_gates() + 6)?;
```
vlopes11, line 301: As @moCello mentioned, we should avoid magic numbers. Since its used in different parts, consider creating a documented constant

src/circuit.rs
```diff
@@ -306,7 +308,7 @@ where
 
         // Add ProverKey to Prover
         prover.prover_key = Some(prover_key.clone());
-        prover.prove(&ck)
+        prover.prove(&ck, &mut rng)
```
vlopes11, line 311: We take `rng` as `mut rng: &mut R`. Mutable references are copy, even if the referred value isn't copy (they are just pointers).
```suggestion
        prover.prove(&ck, rng)
```

The current code works because `&mut R` is virtually congruent to `&mut &mut R`, but I believe this is a nit. We shouldn't take the reference as mutable, since we don't change the reference itself, but the data it points to (RNG).

src/circuit.rs
```diff
@@ -290,13 +291,14 @@ where
 
     /// Generates a proof using the provided `CircuitInputs` & `ProverKey`
     /// instances.
-    fn prove(
+    fn prove<R: RngCore + CryptoRng>(
         &mut self,
         pub_params: &PublicParameters,
         prover_key: &ProverKey,
         transcript_init: &'static [u8],
+        mut rng: &mut R,
```
vlopes11, line 299: Check below
```suggestion
        rng: &mut R,
```

src/permutation.rs
vlopes11: Agreed with @moCello

src/proof_system/preprocess.rs
vlopes11: I'm also not sure how this is related to the blinding scalars. Can we document somewhere (issue?) why we need to do this in order to add blinding scalars?

src/proof_system/prover.rs
```diff
@@ -145,6 +147,30 @@ impl Prover {
...
+        w_vec: &Vec<dusk_bls12_381::BlsScalar>,
+        hiding_degree: usize,
+        domain: &EvaluationDomain,
+        mut rng: &mut R,
```
vlopes11, line 159: Ditto; mutable references are copy

vlopes11 (changes requested):
Few nits/questions

src/permutation.rs
xevisalle: the function `domain.ifft(w_vec)` requires a vector, we could compute it into the function but then in the prover, we need to treat each vector to be blinded in a different way. I think is more clear and clean to put the IFFT into the `blind_poly()`, and treat all inputs in the same way.

src/proof_system/preprocess.rs
xevisalle: we are adding these values, which increases the size of the polynomial by 2 degrees. This means that the polynomial used before no longer fits into `4n`, so, the next power of 2 is `8n`.

src/proof_system/prover.rs
xevisalle: who blinds the polynomials is the prover, so I guess that the best option is to be a method of the prover

src/proof_system/prover.rs
xevisalle: thanks! :)

src/proof_system/prover.rs
xevisalle: I just followed the same naming used in the whole code, we should rename the whole code if we want to change it here... And I used `w_vec_i` where `i` comes from `inverse`. I used this name the distinguish the value we get before calling `Polynomial::from_coefficients_vec()`

src/proof_system/prover.rs
moCello: Hmm, I'm not sure if I agree. I would say the method should be added to the type it operates on. The same way I also wouldn't add a `drive` method to the `Driver` but to the `Car` to give an example.

src/proof_system/prover.rs
moCello: Oh i see, then please call it `w_vec_inverse`. At first sight I thought that you used the `i` because you want to loop through the coefficients later on

src/proof_system/preprocess.rs
vlopes11: As side note, we shouldn't add comments that depends on historical context to the static code. Instead, we need to put self-resolving complete information.

Something like "8n domain, being 4n *add here some PLONK context*, xn for the blinders, and the remainder to fit the next power of 2`

vlopes11 (commented):
Small non-blocking question; waiting for the comments of @moCello to be addressed in order to approve

src/fft/polynomial.rs
```diff
@@ -131,6 +132,31 @@ impl Polynomial {
 
         Ok(Polynomial { coeffs })
     }
+
+    /// Adds the blinding scalars to a given vector. Always the same elements
+    /// of 'w_vec' are modified at the beginning of it, and appended at the end:
+    /// if hiding degree = 1: (b2*X(n+1) + b1*X^n - b2*X - b1) + w_vec
+    /// if hiding degree = 2: (b3*X^(n+2) + b2*X(n+1) + b1*X^n - b3*X^2 - b2*X
```
moCello, line 139: missing the `^` before the `(n+1)` in both lines

moCello (commented):
Find the last approach better than this one. I think the decision we need to make is, if we want to perform the ifft before calling `blind` and have `blind` in part of the `Polynomial` or if we perform the ifft in `blind` and have `blind` part of the `Prover` as you had before. Your call :)

vlopes11 (approved):
LGTM

This is an important security fix and we might consider including this in testnet. cc @autholykos

xevisalle (approved):

Issue #661, opened 2022-02-15
Optimize `append_output_witness`


moCello:
**Describe what you want implemented** 
Try to avoid using `BlsScalar::invert()` in `TurboComposer::append_output_witness`.

**Describe "Why" this is needed**
The `invert` function is expensive as it requires reducing the Scalar to its non-Montgomery form. Since we invert `q_o` and since `q_o` is most of the time either `1` or `-1` (both numbers being the inverse of themselves), we can avoid the `invert` function in these cases.

**Describe alternatives you've considered**
NA

**Additional context**
This has a huge effect on performance of circuit creation (e.g. the zelbet proof creation time is expected to be ~80% faster)

moCello:
issue moved to dusk-bls12_381 https://github.com/dusk-network/bls12_381/issues/78

PR #662, opened 2022-02-15
Optimize performance of append_output_witness

moCello:
Resolves: #661

### DIFF
--- a/src/constraint_system/composer.rs
+++ b/src/constraint_system/composer.rs
@@ -576,6 +576,8 @@ impl TurboComposer {
         self.n += 1;
     }
 
+    /// Calculate the output witness `o` of a [`Constraint`] and append it.
+    /// `o = (q_m · a · b  + q_l · a + q_r · b + q_4 · d + q_c + PI) / (-q_m)`
     pub(crate) fn append_output_witness(&mut self, s: Constraint) -> Witness {
         let a = s.witness(WiredWitness::A);
         let b = s.witness(WiredWitness::B);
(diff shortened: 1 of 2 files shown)


src/constraint_system/composer.rs
```diff
@@ -585,20 +587,23 @@ impl TurboComposer {
...
         let pi = s.coeff(Selector::PublicInput);
 
-        let x = qm * a * b + ql * a + qr * b + qd * d + qc + pi;
-
-        let y = s.coeff(Selector::Output);
-        let y: Option<BlsScalar> = y.invert().into();
-        let y = -y.expect("Inconsistent internal usage of `Constraint::evaluate_output`: Output selector should not be zero");
+        let dividend = q_m * a * b + q_l * a + q_r * b + q_d * d + q_c + pi;
+        let q_o: BlsScalar = *s.coeff(Selector::Output);
+        let divisor_inverse = if q_o * q_o == BlsScalar::one() {
```
vlopes11, line 599: Something sounds off. We perform an extra scalar multiplication for no clear reason when we could check individually for either `-1` or `1`.

Also, we have a clear tech debt in that `expect`: a malicious user can crash a plonk server by submitting a wrong proof. If we are to refactor this part, we should either change the signature of the function to `pub(crate) fn append_output_witness(&mut self, s: Constraint) -> Witness`, or set a constant when `q_o` happens to be zero.

Changing the signature is preferable because it will reflect what happens internally: any proof request with `q_o == 0` is invalid

vlopes11 (changes requested):

src/constraint_system/composer.rs
moCello: I agree with the that the extra scalar multiplication is not needed here, will check for `1` and `-1` individually.

Not sure what you mean with the change of the signature? I assume you mean that we return a `Result<Witness, Error>` ?
I would also avoid setting q_o to `0` manually since it should not be `0` at this point and if it is, it might point to an error somewhere else.

moCello: Thinking of it, why do we actually not check for `1` and `-1` directly in `invert()`? Might give us some performance increase at other spots too

vlopes11: Its quite easy to have `q_o` as `0` here. The user just have to not set anything in the output wire.

moCello: This moved to [bls12_381](https://github.com/dusk-network/bls12_381/issues/78)

moCello:
The check for `1` and `-1` moved to [bls12_381](https://github.com/dusk-network/bls12_381/issues/78)

And the tech debt with the `expect` will be addressed in its own issue #670

PR #677, opened 2022-04-06
Add the PIs to the transcript

xevisalle:


src/proof_system/verifier.rs
```diff
@@ -89,6 +90,12 @@ impl Verifier {
         public_inputs: &[BlsScalar],
     ) -> Result<(), Error> {
         let mut cloned_transcript = self.preprocessed_transcript.clone();
+
+        // PIs have to be part of the transcript
+        for pi in public_inputs.iter() {
+            cloned_transcript.append_scalar(b"pi", pi);
+        }
```
vlopes11, line 97: We have a discrepancy between the prover and verifier here.

In the prover, we are using the dense public inputs (that is just the expected public inputs filled with zeroes to match their expected indices).

Here we are using just a not indexed small set of public inputs. Is that intended? If yes, why is that correct?

vlopes11: Sounds good, we create a dense PI before coming to this part:

https://github.com/dusk-network/plonk/blob/c6af25f6693e9e06a259a87f321e7dd2298b5c57/src/circuit.rs#L344-L372

src/proof_system/prover.rs
```diff
@@ -197,6 +197,17 @@ impl Prover {
         // the Transcript has been seeded with the preprocessed commitments
         let mut transcript = self.preprocessed_transcript.clone();
 
+        // PIs have to be part of the transcript
+        for pi in self.cs.to_dense_public_inputs().iter() {
+            transcript.append_scalar(b"pi", pi);
+        }
+
+        // We fill with zeros up to the domain size, just like the verifier will
+        // do
```
moCello, line 206: I find this comment confusing since it is not immediately clear to me where the verifier will fill the `transcript` with zeros

moCello (approved):
LGTM, just that I am still a bit confused regarding the comment where the zeros are appended to the `transcript`. If possible it would be great if you can clarify the comment :)

PR #685, opened 2022-05-23
Rebase to original Plonk

xevisalle:
This PR rebases the code to the original Plonk implementation, but keeping all the recent improvements. It does the following:
- Modify the prover and the verifier according to the original Plonk.
- Remove all the Plonkup related code.


moCello (approved):
As far as I can tell it LGTM, just one comment that you might want to add but that is up  to you

PR #687, opened 2022-05-24
Fix verification time

xevisalle:
The verification time was growing up with the circuit size, because the whole public inputs vector, which is padded with 0s up to the size of the domain, was put all of it into the transcript. We fixed it by putting only the real public inputs.


moCello (approved):
LGTM

PR #691, opened 2022-06-02
Fix error with the max degree of the SRS

xevisalle:
Adding the blinding factors required some extra elements for the SRS: `+1` per each wire (we have 4 wires), plus `+2` for the permutation polynomial. There was a bug in the original commit, that caused errors when compiling some circuits. We fix it here by doing:

- In `srs.rs`, the `max_degree` in `setup()` is updated: `+ 6`
- In `srs.rs`, the truncated degree used in `trim()` is updated: `+ 6`.
- The `+ 6` used in `circuit.rs` has been removed as it is no longer required.
- We reverted a change in `benches/plonk.rs`, no longer required.

### DIFF
--- a/benches/plonk.rs
+++ b/benches/plonk.rs
@@ -81,7 +81,7 @@ fn constraint_system_prove(
 
 fn constraint_system_benchmark(c: &mut Criterion) {
     let initial_degree = 5;
-    let final_degree = 18;
+    let final_degree = 17;
 
     let rng = &mut rand_core::OsRng;
     let label = b"dusk-network";
@@ -89,7 +89,7 @@ fn constraint_system_benchmark(c: &mut Criterion) {
         .expect("Failed to create PP");
 
     let data: Vec<(BenchCircuit, ProverKey, VerifierData, Proof)> =
-        (initial_degree..final_degree)
+        (initial_degree..=final_degree)
             .map(|degree| {
                 let mut circuit = BenchCircuit::from(degree as usize);
                 let (pk, vd) =
--- a/src/commitment_scheme/kzg10/srs.rs
+++ b/src/commitment_scheme/kzg10/srs.rs
@@ -38,19 +38,28 @@ impl PublicParameters {
         &self.opening_key
     }
 
+    /// The maximum degree is the degree of the constraint system + 6,
+    /// because adding the blinding factors requires some extra elements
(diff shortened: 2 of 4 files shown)


benches/plonk.rs
```diff
@@ -85,7 +85,7 @@ fn constraint_system_benchmark(c: &mut Criterion) {
 
     let rng = &mut rand_core::OsRng;
     let label = b"dusk-network";
-    let pp = PublicParameters::setup(1 << final_degree, rng)
+    let pp = PublicParameters::setup(1 << (final_degree - 1), rng)
```
vlopes11, line 88: why is this required for a bug fix?

xevisalle: We are reverting it, this is the original code, that we had to modify in https://github.com/dusk-network/plonk/commit/8efda59b0c07821c4a1ee66f7172143326b28b48 to avoid a crash, related with the same.

src/commitment_scheme/kzg10/srs.rs
```diff
@@ -38,20 +38,29 @@ impl PublicParameters {
         &self.opening_key
     }
 
+    /// The maximum degree is the degree of the constraint system + 6,
+    /// because adding the blinding factors requires some extra elements
+    /// for the SRS: +1 per each wire (we have 4 wires), plus +2 for the
+    /// permutation polynomial
+    const MAX_DEGREE_UPDATE: usize = 6;
```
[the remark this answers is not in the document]
xevisalle: As you prefer!

benches/plonk.rs
xevisalle: Because you are never using so many public parameters, as per how the benchmark loop iterating over all the possible circuit degrees was implemented. The maximum circuit degree benchmark is final degree - 1.

moCello: care to write it as a comment ;)

src/commitment_scheme/kzg10/srs.rs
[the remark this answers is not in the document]
moCello: Sort of curious on your and @vlopes11 opinions here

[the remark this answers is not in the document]
xevisalle: ok I put this then!

moCello (approved):
LGTM

PR #696, opened 2022-06-16
Fix math latex rendering on docs.rs

moCello:
With this PR the math finally should render on docs.rs. 
There is also the make rule `make docs-local` added that opens the docs from the current branch locally.

Resolves: #567

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,4 +7,7 @@ doc: ## Generate documentation
 doc-internal: ## Generate documentation with private items
 	@cargo rustdoc --lib -- --document-private-items -D warnings
 
-.PHONY: help doc doc-internal 
+doc-local: ## Open local documentation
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -7,4 +7,7 @@ doc: ## Generate documentation
 doc-internal: ## Generate documentation with private items
 	@cargo rustdoc --lib -- --document-private-items -D warnings
 
-.PHONY: help doc doc-internal 
+doc-local: ## Open local documentation
+	@RUSTDOCFLAGS="--html-in-header docs/katex-header.html" cargo doc --no-deps --open
+
+.PHONY: help doc doc-internal doc-local
```
vlopes11, line 13: All these `doc-*` are quite confusing. Maybe leave an issue to refactor them into a single document? I don't see why the documentation should be that complicated.

vlopes11 (approved):
LGTM, comment for future improvement

PR #712, opened 2022-10-20
Remove `rkyv` validation feature

ureeves:
This errant manual implementation of `CheckBytes` was using the `rkyv` `validation` feature which bring in `std`. This is not desirable for smart contract development, and would break compilation of the crate with only the `alloc` and `rkyv-impl` features enabled.


ureeves:
This requires this PR to be merged to work:
https://github.com/dusk-network/bls12_381/pull/102

moCello (approved):
LGTM

PR #716, opened 2022-12-05
Lift the `Default` bound for the `Circuit` trait

### LINKED ISSUE #715: `Circuit` shouldn't be required to be `Default`
A circuit shouldn't be required to be `Default` to satisfy the `Circuit` trait. The proposal is to remove said bound, and adjust accordingly.

First this semantically makes no sense - why should a circuit need to have a default to be a circuit? - but it also leads to some dangerous uses.
Some circuits downstream, namely the execute circuits, actually shouldn't have a default implementation since they require padding to be performed prior to their compilation, and compiling with `Compiler::compile` will result in the wrong keys being generated.

N/A

**Additional context**
This shouldn't be a breaking change, as it it a lifting of bounds, rather than a restriction.

ureeves:
The `Compiler::compile` function keeps this bound since it actually requires the circuit to have a `Default` implementation, but all other functionality now works without said bound.

Resolves #715


moCello (approved):
LGTM

vlopes11:
The `Default` trait is meant to be the definition in which the circuit will compile to. The existence of the execute circuit should not be taken as reference because it was created several plonk/rust versions ago, when the language limited for instance associated constants for traits (and a couple of other things).

The general idea behind this concept is that a type should have a single circuit (hence, a keys set). This will provide type-safety and will make it easier to store the keys statically.

vlopes11:
You mean that the user can serialize a PLONK key using a type `A`, and if he deserializes with type `B` this is a violation of type-safety?

If we feel this is indeed a violation to the principle, then my suggestion would be to fix that issue and improve the security instead of removing the concept.

The fact that we can't have this implementation for the execute circuit just highlights it should be refactored to comply with the new rules.

PR #719, opened 2023-01-12
Remove `bit_iterator` macro

### LINKED ISSUE #632: Remove `bit_iterator` macro
Remove the `bit_iterator` macro.

As far as I can see we only invoke the macro once to create `BitIterator8` so I don't see the need to have a macro for it. Creating `BitIterator8` directly is much clearer.

Leave as is.

**Additional context**
NA

ashutoshvarma:
Fix #632

### DIFF
--- a/src/bit_iterator.rs
+++ b/src/bit_iterator.rs
@@ -8,54 +8,47 @@
 //! limbs
 use core::mem;
 
-macro_rules! bit_iterator {
-    ($sty : ty, $name : ident) => {
-        #[derive(Debug, Clone, Copy)]
-        pub struct $name<E> {
-            // scalar is the slice of integers that wish to iterate over
-            scalar: E,
-            // num_of_total_bits represents the sum of all of the bits of each
-            // integer If we have 2 u32s then the total number of bits will
-            // be 32 * 2 = 64 bits
-            num_of_total_bits: usize,
-            // bit_len represents the bit length of each integer.
-            // If we have a slice of u32s, then bit_len will be 32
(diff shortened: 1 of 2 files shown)


ashutoshvarma:
Seems one of the CI checks failed due to an error in uploading a coverage report to codecov.

src/bit_iterator.rs
```diff
@@ -8,54 +8,47 @@
...
-            scalar: E,
-            // num_of_total_bits represents the sum of all of the bits of each
-            // integer If we have 2 u32s then the total number of bits will
-            // be 32 * 2 = 64 bits
-            num_of_total_bits: usize,
-            // bit_len represents the bit length of each integer.
-            // If we have a slice of u32s, then bit_len will be 32
-            bit_len: usize,
-        }
+pub struct BitIterator8<E> {
```
moCello, line 11: You didn't derive the `Debug`, `Clone` and `Copy` traits. Can you still include those?

moCello (changes requested):
Very nice, thanks!
Please also include the `Debug`, `Clone` and `Copy` traits as mentioned in the comments.

And it would be great if you could also add a section in the CHANGELOG under 'Unreleased', like:

### Removed

- Remove `bit_iterator` macro [#632]

And link the issue at the bottom of the CHANGELOG

src/bit_iterator.rs
ashutoshvarma: done!

moCello (approved):
LGTM

Issue #720, opened 2023-01-18
Bug in point equality assertion


ilitteri:
**Describe the bug**
Point equality is not being done correctly. `b`'s `y` component is being compared to itself [here](https://github.com/dusk-network/plonk/blob/master/src/composer.rs#L635) 

**To Reproduce**
1. `rustup update nightly`
2. `rustup override set nightly`
3. Run the test above (`p1`'s `y` component is different than `p2`'s `y` component and the proving fails.

```rust
#[test]
fn failing_test() {
    let rng = &mut StdRng::seed_from_u64(8349u64);

    let n = 1 << 12;
    let label = b"demo";
    let pp = PublicParameters::setup(n, rng).expect("failed to create pp");

    pub struct DummyCircuit {
        p1: JubJubExtended,
        p2: JubJubExtended,
    }

    impl Default for DummyCircuit {
        fn default() -> Self {
            Self {
                p1: JubJubExtended::from_raw_unchecked(
                    BlsScalar::zero(),
                    BlsScalar::one(),
                    BlsScalar::one(),
                    BlsScalar::zero(),
                    BlsScalar::zero(),
                ),
                p2: JubJubExtended::from_raw_unchecked(
                    BlsScalar::zero(),
                    BlsScalar::zero(),
                    BlsScalar::one(),
                    BlsScalar::zero(),
                    BlsScalar::zero(),
                ),
            }
        }
    }

    impl Circuit for DummyCircuit {
        fn circuit<C>(&self, composer: &mut C) -> Result<(), Error>
        where
            C: Composer,
        {
            let w_x = composer.append_point(self.p1);
            let w_y = composer.append_point(self.p2);
            composer.assert_equal_point(w_x, w_y);
            Ok(())
        }
    }

    let (prover, _verifier) = Compiler::compile::<DummyCircuit>(&pp, label)
        .expect("failed to compile circuit");

    let proving_result = prover
        .prove(rng, &Default::default());

    assert!(proving_result.is_err(), "proving should fail because the points are not equal");
}
```

**Expected behavior**
Proving should fail because the points that are being compared are not equal.

**Logs/Screenshot**
```sh
thread 'failing_test' panicked at 'proving should fail because the points are not equal', tests/composer.rs:167:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4781233a77e879e49cb5ce3c98d2abba6a6ade7a/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
[...]

PR #721, opened 2023-01-18
Fix point equality assertion

ilitteri:
Solves #720

### DIFF
--- a/src/composer.rs
+++ b/src/composer.rs
@@ -632,7 +632,7 @@ pub trait Composer: Sized + Index<Witness, Output = BlsScalar> {
     /// Asserts `a == b` by appending two gates
     fn assert_equal_point(&mut self, a: WitnessPoint, b: WitnessPoint) {
         self.assert_equal(*a.x(), *b.x());
-        self.assert_equal(*b.y(), *b.y());
+        self.assert_equal(*a.y(), *b.y());
     }
 
     /// Asserts `point == public`.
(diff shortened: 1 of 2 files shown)


moCello (changes requested):
Well spotted, thanks! Just one formatting error

tests/composer.rs
```diff
@@ -111,3 +111,58 @@ fn circuit_with_all_gates() {
...
+        }
+    }
+
+    let (prover, _verifier) = Compiler::compile::<DummyCircuit>(&pp, label)
+        .expect("failed to compile circuit");
+
+    let proving_result = prover
+        .prove(rng, &Default::default());
+
+    assert!(proving_result.is_err(), "proving should fail because the points are not equal");
```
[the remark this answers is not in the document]
ilitteri: Sorry, I forgot to do `cargo fmt` 😅, it should be fixed now

moCello (approved):
LGTM

src/composer.rs
```diff
@@ -632,7 +632,7 @@ pub trait Composer: Sized + Index<Witness, Output = BlsScalar> {
     /// Asserts `a == b` by appending two gates
     fn assert_equal_point(&mut self, a: WitnessPoint, b: WitnessPoint) {
         self.assert_equal(*a.x(), *b.x());
-        self.assert_equal(*b.y(), *b.y());
```
mimoo, line 635: you can avoid these kind of mistakes by deconstructing btw, so that the compiler yells at you if you forget a variable. For example:

```rust
fn assert_equal_point(&mut self, a: WitnessPoint, b: WitnessPoint) {
    let (x1, y1) = a.coords();
    let (x2, y2) = b.coords();
    self.assert_equal(x1, x2);
    self.assert_equal(y1, y2);
}
```

moCello: unfortunately `a` and `b` are `WitnessPoint`s and don't have a `coord` method.

mimoo: you can implement one, or make x, y public :o

PR #723, opened 2023-01-19
Remove 'src/circuit.rs' file

### LINKED ISSUE #722: Remove 'circuit.rs' file
Remove 'src/circuit.rs' file since the module now exists under the 'composer' module but the original file seems to not have been deleted when the module moved.

moCello:
The `circuit` module moved into the `composer` module but the original file was still around while not included in the module tree.

Resolves #722


ureeves (approved):
LGTM

PR #726, opened 2023-01-24
Add more tests for `assert_equal_point`

### LINKED ISSUE #725: Add more tests for point equality assertion
Add tests for `assert_equal_point` method of the composer:
- sanity test
- fails when the points are not equal
- fails when only one of the coordinates are not equal

We need to increase our tests coverage, as the issue #721 clearly showed.

N/A

**Additional context**
N/A

moCello:
Resolves #725


tests/ecc.rs
```diff
@@ -304,3 +304,202 @@ fn mul_point_works() {
             .expect_err("circuit is not satisfied");
     }
 }
+
+#[test]
+fn assert_equal_point_works() {
```
xevisalle, line 309: I suggest having only one circuit that executes the assertation, and the function is called with two points as arguments. Then, compute each variant in each test and call that circuit.

tests/ecc.rs
```diff
@@ -304,3 +304,202 @@ fn mul_point_works() {
...
+        fn circuit<C>(&self, composer: &mut C) -> Result<(), Error>
+        where
+            C: Composer,
+        {
+            // test: generator = generator
+            let w_p1 = composer.append_point(self.point);
+            let w_p2 = composer.append_point(self.point);
+            composer.assert_equal_point(w_p1, w_p2);
+
+            // test: 42 * generator = 42 * generator
```
xevisalle, line 338: personally, I'd prefer to hardcode one of the points and compute the other one. But that's a nit.

tests/ecc.rs
moCello: Yeah, I'm actually just doing these changes! Was a bit too quick setting the changes as done, sorry!

xevisalle (approved):

PR #732, opened 2023-01-30
Refactor integration tests

moCello:
This issue covers:
- add helper functions:
  - `setup`
  - `check_satisfied_circuit`
  - `check_unsatisfied_circuit`
- refactor tests:
  - `assert_equal_point`
  - `assert_equal_public_point`
  - `component_mul_point`
  - `component_add_point`
  - `component_mul_generator`
  - `component_boolean`
  - `component_select`
  - `component_select_one`
  - `component_select_zero`
  - `component_select_point`
  - `component_select_identity`

Resolves #731

PR #739, opened 2023-02-07
Add tests for `assert_equal` and `assert_equal_constant`

### LINKED ISSUE #733: Add tests for `assert_equal` and `assert_equal_constant`
Add in-circuit tests for the composer methods
- `assert_equal`
- `assert_equal_constant`

Increase test coverage

N/A

**Additional context**
Continuation of #731

moCello:
Resolves #733


tests/select_point.rs
```diff
@@ -0,0 +1,359 @@
...
+            let w_result = composer.append_point(self.result);
+
+            let result_circuit =
+                composer.component_select_point(w_bit, w_point_a, w_point_b);
+
+            composer.assert_equal_point(w_result, result_circuit);
+
+            Ok(())
+        }
+    }
```
ureeves, line 67: Should we, for the sake of example completeness, constrait the `bit` to be 0 or 1?

tests/common/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+// Check that proof creation and verification of a satisfied circuit fails
+// when the public inputs from the test circuit does not match the ones from
+// the verifier circuit description
+// As this is a very specific test case, this function will not be used by all
+// tests.
+#[allow(dead_code)]
+pub(crate) fn check_satisfied_circuit_fails<C, R>(
+    prover: &Prover<C>,
+    verifier: &Verifier<C>,
+    pi_expected: &Vec<BlsScalar>,
```
ureeves, line 57: Just in general taking a reference to a vector can be easily substituted by a slice to the items in the vector.

ureeves (approved):
One nit, and one circuit question, but not enough to block merging. LGTM

tests/select_point.rs
moCello: You're completely right!

tests/common/mod.rs
moCello: For the tests I'll leave it as is if you don't mind

tests/select_point.rs
moCello: On second thought though: I actually added tests that pass even though `bit` is not 1 or 0 to specifically test that behavior. So I might opt to add a comment, that clarifies that any `component_select` should _always_ constrain `bit` _before_ the component is called.

moCello: In any case: the `component_select*` tests were part of the last issue #732 so I'll go ahead and merge this one

PR #744, opened 2023-02-27
Fix `ProverKey::serialization_size`

### LINKED ISSUE #743: Fix `ProverKey::serialization_size`
With the change from plonkup back to plonk #684 in the serialization size of the proverkey the `poly_num` were not adapted correctly, causing the serialization size to be bigger than needed.
Additionally the number of coefficients in a constraint also weren't updated. There used to be 13 coefficients but now that the lookupgate is gone, there are only 12 coefficients in a constraint.

The serialization size was bigger than it needed to be. While the other way round would have much worse implications, this is still to be fixed.

N/A

**Additional context**
N/A

moCello:
Resolves #743


ureeves (approved):
LGTM

PR #748, opened 2023-03-14
Fix range gate

### LINKED ISSUE #746: No range check when `num_bits = 0`
**Describe the bug**
When adding a range gate with zero bits, any witness always satisfies the circuit.

**To Reproduce**
In 'tests'
```rust
use dusk_plonk::prelude::*;
use rand::rngs::StdRng;
use rand::SeedableRng;

mod common;
use common::{check_satisfied_circuit, check_unsatisfied_circuit, setup};

#[test]
fn range() {
    #[derive(Default)]
    pub struct TestCircuit {
        a: BlsScalar,
        bits: usize,
    }

    impl TestCircuit {
        pub fn new(a: BlsScalar, bits: usize) -> Self {
            Self { a, bits }
        }
    }

    impl Circuit for TestCircuit {
        fn circuit<C>(&self, composer: &mut C) -> Result<(), Error>
        where
            C: Composer,
        {
            let w_a = composer.append_witness(self.a);

            composer.component_range(w_a, self.bits);

            Ok(())
        }
    }

    // Compile common circuit descriptions for the prover and verifier to be
    // used by all tests
    let label = b"component_range";
    let rng = &mut StdRng::seed_from_u64(0xb1eeb);
    let capacity = 1 << 6;
    let (prover, verifier) =
        setup(capacity, rng, label, &TestCircuit::default());

    // public input to be used by all tests
    let pi = vec![];

    // Test bits = zero
    //
    // Test default works:
    // 0 < 2^0
    // This is an edge case but one can argue that this case can pass
    let msg = "Default circuit verification should pass";
    let circuit = TestCircuit::default();
    check_satisfied_circuit(&prover, &verifier, &pi, &circuit, rng, &msg);

    // Test:
    // 1 !< 2^0
    // In this case it shouldn't be possible to generate a circuit that is satisfied
    let msg = "Verification of satisfied circuit should pass";
    let bits = 0;
    let a = BlsScalar::one();
    let circuit = TestCircuit::new(a, bits);
    check_unsatisfied_circuit(&prover, &circuit, rng, &msg);

    // Test:
    // random !< 2^0
    // In this case it shouldn't be possible to generate a circuit that is satisfied
    let msg = "Unsatisfied circuit should fail";
    let a = BlsScalar::random(rng);
    assert!(a != BlsScalar::zero());
    let circuit = TestCircuit::new(a, bits);
    check_unsatisfied_circuit(&prover, &circuit, rng, &msg);
}
```

**Expected behaviour**
No witness should satisfy a range gate with 0 bits (since no number can be encoded in 0 bits). An edge case would be `BlsScalar::zero()` itself as one can argue that zero is encodable in 0 bits.

**Additional context**
This bug came up in issue #735 and the above test is in 'tests/range.rs'

moCello:
Resolves #746

### DIFF
--- a/src/composer.rs
+++ b/src/composer.rs
@@ -910,8 +910,8 @@ pub trait Composer: Sized + Index<Witness, Output = BlsScalar> {
     /// Adds a range-constraint gate that checks and constrains a [`Witness`]
     /// to be encoded in at most `num_bits`.
     ///
-    /// This function adds `num_bits/4` gates to the circuit description in
-    /// order to add the range constraint.
+    /// This function adds min(1, `num_bits/4`) gates to the circuit description
+    /// in order to add the range constraint.
     ///
     ///# Panics
     /// This function will panic if the num_bits specified is not even, ie.
@@ -919,6 +919,13 @@ pub trait Composer: Sized + Index<Witness, Output = BlsScalar> {
     fn component_range(&mut self, witness: Witness, num_bits: usize) {
         // number of bits must be even
         debug_assert_eq!(num_bits % 2, 0);
+
+        // if num_bits = 0 constrain witness to 0
+        if num_bits == 0 {
+            let constraint = Constraint::new().left(1).a(witness);
+            self.append_gate(constraint);
+            return;
+        }
 
         // convert witness to bit representation and reverse
         let bits = self[witness];
--- a/src/composer.rs
+++ b/src/composer.rs
@@ -908,7 +908,8 @@ pub trait Composer: Sized + Index<Witness, Output = BlsScalar> {
     }
 
     /// Adds a range-constraint gate that checks and constrains a [`Witness`]
-    /// to be encoded in at most `num_bits`.
+    /// to be encoded in at most `num_bits`, which means that it will be within
+    /// the range `[0, 2^num_bits[`.
     ///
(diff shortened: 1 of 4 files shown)


src/composer.rs
```diff
@@ -907,11 +907,11 @@ pub trait Composer: Sized + Index<Witness, Output = BlsScalar> {
         self.gate_mul(constraint)
     }
 
-    /// Adds a range-constraint gate that checks and constrains a
-    /// [`Witness`] to be inside of the range \[0,num_bits\].
+    /// Adds a range-constraint gate that checks and constrains a [`Witness`]
+    /// to be encoded in at most `num_bits`.
```
ureeves, line 911: Perhaps putting in the actual number range would be a good idea? Not like it was before, but something like `[0, 2^num_nbits[`

ureeves (approved):
LGTM, but one comment.

src/composer.rs
moCello: I myself found the range in the comment slightly confusing as it gave me the impression that the number itself is in that range.
What this gate is actually doing though is constraining a number to be encoded in at most `num_bits` which will make it be in the range `[0, 2^num_bits[`. Do you want me to add this last part?

ureeves: That's what I thought it was doing. It's good to clarify this.

Issue #763, opened 2023-08-14
Tests fail when run in debug mode


moCello:
**Describe the bug**
All our tests pass when run in `release` mode but some fail when run in `debug` mode, notably the example from the readme, `append_logic_xor` and `append_logic_and`.

**To Reproduce**
For the example from the README run
```
cargo test --doc
```
and for the logic (and possibly other) tests run
```
cargo test
```

**Expected behaviour**
The tests should pass no matter if in `debug` or `release` mode

**Version**
v0.14.1

**Additional context**
First discovered in #753 and #758

PR #764, opened 2023-08-16
Fix tests panic in debug

moCello:
Resolves: #763

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ use rand_core::OsRng;
 // Implement a circuit that checks:
 // 1) a + b = c where C is a PI
 // 2) a < 2^6
-// 3) b < 2^5
+// 3) b < 2^4
 // 4) a * b = d where D is a PI
 // 5) JubJub::GENERATOR * e(JubJubScalar) = f where F is a Public Input
 #[derive(Debug, Default)]
@@ -50,8 +50,10 @@ impl Circuit for TestCircuit {
         composer.append_gate(constraint);
 
         // Check that a and b are in range
-        composer.component_range(a, 6);
-        composer.component_range(b, 5);
+        const HALF_SIX: usize = 3;
+        composer.component_range::<HALF_SIX>(a);
+        const HALF_FOUR: usize = 2;
+        composer.component_range::<HALF_FOUR>(b);
 
         // Make second constraint a * b = d
         let constraint =
--- a/src/composer.rs
+++ b/src/composer.rs
@@ -105,10 +105,10 @@ pub trait Composer: Sized + Index<Witness, Output = BlsScalar> {
         self.append_custom_gate_internal(constraint)
     }
 
-    /// Performs a logical AND or XOR op between the inputs provided for the
-    /// specified number of bits (counting from the least significant bit).
+    /// Performs a logical AND or XOR op between the inputs provided for
+    /// `num_bits = BIT_PAIRS * 2` bits (counting from the least significant).
     ///
-    /// Each logic gate adds `(num_bits / 2) + 1` gates to the circuit to
+    /// Each logic gate adds `BIT_PAIRS + 1` gates to the circuit to
     /// perform the whole operation.
     ///
     /// ## Constraint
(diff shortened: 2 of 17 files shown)


src/composer.rs
```diff
@@ -129,7 +129,7 @@ pub trait Composer: Sized + Index<Witness, Output = BlsScalar> {
     ) -> Witness {
         // the bits are iterated as chunks of two; hence, we require an even
         // number
-        debug_assert_eq!(num_bits & 1, 0);
+        assert_eq!(num_bits & 1, 0, "number of bits must be even");
```
vlopes11, line 132: Not sure if it makes sense to add production panic to check circuit development consistency. I'd rather keep it as `debug_assert_eq` and have tests checking for such case

src/composer.rs
```diff
@@ -913,18 +919,19 @@ pub trait Composer: Sized + Index<Witness, Output = BlsScalar> {
...
     ///# Panics
-    /// This function will panic if the num_bits specified is not even, ie.
-    /// `num_bits % 2 != 0`.
+    /// This function will panic if the num_bits specified is not even since
+    /// that results in potentially incorrect circuits:
+    /// A circuit constraining 63 < 2^5 for example would validate to true.
     fn component_range(&mut self, witness: Witness, num_bits: usize) {
-        // number of bits must be even
-        debug_assert_eq!(num_bits % 2, 0);
+        assert_eq!(num_bits % 2, 0, "number of bits must be even");
```
vlopes11, line 934: Consider using the same assert as before: `num_bits & 1, 0`

vlopes11 (commented):
Looks good overall! A comment regarding the correct change on enforcing range checks of even values

src/composer.rs
ureeves: I have the same feeling on panicking like this, but more because it signals that we should be erroring rather than panicking. The library should validate user input against what it expects.

ureeves (changes requested):
The instinct to want to actually panic on this makes sense, but I think it makes more sense to error in this situation. Here's the book on [when to panic](https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html).

I'm open to being persuaded otherwise though.

src/composer.rs
vlopes11: There is an important distinction tho: the validation is static and should be performed at compilation time.

Meaning: when we create a circuit implementation, we create it for a static type (i.e. `impl Circuit for Foo`). It means it could be a waste to move such check to runtime (i.e. make the function fallible & `Err`).

If we move everything to static time, then it probably makes more sense. In short: if the function receives a constant, we can just `assert` it and it would fail at compile time if the user attempts to create a circuit with wrong bits boundaries:

```rust
fn append_logic_component<BITS: usize>(
        &mut self,
        a: Witness,
        b: Witness,
        is_component_xor: bool,
    ) -> Witness {
    assert_eq!(BITS & 1, 0, "some msg");
```

The code above will not compile for invalid bits.

vlopes11: In fact, the problem above probably applies elsewhere. We should probably use more static stuff for other components instead of running test checks. Probably `is_component_xor` should also be a constant.

TLDR; components should receive only witnesses as arguments. Everything else is a constant

moCello: We can also circumvent any misuse if we let the user pass the amount of bit-pairs instead of bits. In the components themselves we multiply the bit-pairs by 2 and can therefore always be sure the total amount of bits is dividable by two. 

It would be more cognitive overhead on the circuit creator but would make sure the components can not be misused without the need to panic or error

moCello: I did an attempt on my above proposal combined with @vlopes11 suggestion of making the logic and range components generic over the bit (bit-halves in this case).

@vlopes11 and @ureeves please check it out under the branch: [mocello/763_fix_debug_panic_bit_pairs](https://github.com/dusk-network/plonk/tree/mocello/763_fix_debug_panic_bit_pairs) and let me know whether you like that one or the latest changes on this branch more.

ureeves: I very much like the direction of "making illegal state unrepresentable". Comes with tons of benefits. I vote for the [`bit_pairs`](https://github.com/dusk-network/plonk/tree/mocello/763_fix_debug_panic_bit_pairs) branch that you suggested.

moCello: I also like the solution with the `BIT_PAIRS` most.

Pushed to the branch, please have a final look and approve (*if* you approve that is ;) )

ureeves (approved):
LGTM

README.md
```diff
@@ -50,8 +50,10 @@ impl Circuit for TestCircuit {
         composer.append_gate(constraint);
 
         // Check that a and b are in range
-        composer.component_range(a, 6);
-        composer.component_range(b, 5);
+        const HALF_SIX: usize = 3;
```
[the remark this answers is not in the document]
moCello: The range (and logic) component is now generic over a constant `BIT_PAIRS`. This was done so that the user has no way to construct an invalid circuit (internally we group the bits in pairs which made circuits over an uneven number of bits invalid).

In the case of the above example, we chose `3` for the generic `BIT_PAIRS`, which will result in a circuit that checks whether the witness is encoded in 6 bits or less.
To make that connection between 3 and 6 clear I chose the name `HALF_SIX`.

PR #780, opened 2023-10-18
Fix append_logic_xor test

xevisalle:
Fixes a problem in the tests related to `append_logic_xor`. It computed an xor operation without taking into account the modulus applied to all operations. Basically:

`((a ^ b) & bit_mask) % mod != (((a ^ b) % mod) & bit_mask) % mod`


moCello (approved):
LGTM

PR #783, opened 2023-10-25
refactor polynomial operations

ashWhiteHat:
I refactored polynomial operations because it was a little bit of complicated.
The reason why I removed `interpolate` was that I would like to work on removal of `Evaluations::domain` as follows and `interpolate` would be the bottleneck of it.
https://github.com/dusk-network/plonk/blob/master/src/fft/evaluations.rs#L39

I would appreciate it if you could confirm.
Thank you.

### DIFF
--- a/src/fft/evaluations.rs
+++ b/src/fft/evaluations.rs
@@ -7,7 +7,6 @@
 //! A polynomial represented in evaluations form over a domain of size 2^n.
 
 use super::domain::EvaluationDomain;
-use super::polynomial::Polynomial;
 use crate::error::Error;
 use alloc::vec::Vec;
 use core::ops::{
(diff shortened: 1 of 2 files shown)


ashWhiteHat:
I also think that we should integrate `fft::evaluations.rs` and `fft::polynomial.rs`.
Polynomial has two representation **Coefficient** and **Point-Value**.
`fft::evaluations.rs` is **Point-Value** form of Polynomial thus it should be in `fft::polynomial.rs`.
`Evaluations::domain` is only used for [domain equation](https://github.com/dusk-network/plonk/blob/master/src/fft/evaluations.rs#L129) so we can remove it.
Then, `fft::evaluations.rs` has a few code so we can move it to `fft::polynomial.rs`.

Polynomial Representation
https://web.cecs.pdx.edu/~maier/cs584/Lectures/lect07b-11-MG.pdf

PR #785, opened 2023-10-27
Add JubJubScalarMalformed error

### LINKED ISSUE #784: Add `JubJubScalarMalformed` error
#### Summary
in `component_mul_point` we through an `BlsScalarMalformed` error when the `JubJubScalar` witness is malformed. This is wrong and a new error type for this case needs to be created.

#### Relevant Context
see https://github.com/dusk-network/schnorr/pull/88#discussion_r1374292888

HDauven:
Resolves #784

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -5,6 +5,15 @@ coverage:
   precision: 2
   round: down
   range: "70...100"
+  status:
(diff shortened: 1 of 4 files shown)


moCello (approved):
LGTM

PR #800, opened 2023-12-14
Fix leading coefficients might be zero

### LINKED ISSUE #796: `leading_coefficient()` may unexpectedly be zero
#### Summary
Polynomials are stored as a vector of `BlsScalars`. A polynomial is the zero polynomial if 1) it is empty or 2) all coefficients in the vector are zero. We don’t want to store extra zero coefficients if we don’t need to, so almost every function on polynomials makes sure to call `self.truncate_leading_zeros()` after each operation.

However, there are two exceptions: in `add_assign` and in `sub_assign`, if two polynomials have the same degree, then `truncate_leading_zeros()` is not called. It’s possible for the highest coefficients to cancel out, leaving us with a non-zero polynomial that still has a “zero” leading coefficient.

In 'fft/polynomial.rs', the `leading_coefficient()` function should return the largest non-zero coefficient of the polynomial. Currently, this function is marked as dead_code and not used anywhere in the repository, but future developers of this library could easily make a mistake here.
Recommendation: Either fix `leading_coefficient()` to ignore leading zeros, or change `add_assign` and `sub_assign` to call `truncate_leading_zeros` in the case of two polynomials with the same degree.

#### Relevant Context (optional)
Finding of the plonk audit.

moCello:
Resolves #796

### DIFF
--- a/src/fft/polynomial.rs
+++ b/src/fft/polynomial.rs
@@ -78,8 +78,7 @@ impl Polynomial {
     /// When the length of the coeffs is zero.
     pub(crate) fn from_coefficients_vec(coeffs: Vec<BlsScalar>) -> Self {
         let mut result = Self { coeffs };
-        // While there are zeros at the end of the coefficient vector, pop them
-        // off.
+        // If the leading coefficients end up being zero, pop them off.
         result.truncate_leading_zeros();
         // Check that either the coefficients vec is empty or that the last
         // coeff is non-zero.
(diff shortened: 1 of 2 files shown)


src/fft/polynomial.rs
```diff
@@ -113,30 +118,35 @@ impl Polynomial {
         }
     }
 
-    /// Evaluates a [`Polynomial`] at a given point in the field.
-    pub(crate) fn evaluate(&self, point: &BlsScalar) -> BlsScalar {
+    /// Evaluates a [`Polynomial`] at a given scalar in the field.
```
xevisalle, line 121: even when is the common way to say it, indeed was a bit confusing to talk about 'points' in this context, as it could seem that it refers to a point in the curve. However, I would not put 'scalar' here as it is not a scalar (even when the type is). I would say 'value'.

src/fft/polynomial.rs
```diff
@@ -91,16 +90,22 @@ impl Polynomial {
         result
     }
 
-    /// Returns the degree of the [`Polynomial`].
+    /// Returns the degree, i.e. the index of the highest non-zero coefficient,
```
xevisalle, line 93: I understand what you say, but could it seem that you are taking the highest coefficient? Maybe say the highest index among all coeffs...?

HDauven (approved):
LGTM!

src/fft/polynomial.rs
```diff
@@ -192,6 +206,7 @@ impl<'a, 'b> Add<&'a Polynomial> for &'b Polynomial {
             }
             result
         };
+        // If the leading coefficients end up being zero, pop them off.
         result.truncate_leading_zeros();
```
xevisalle, line 210: just noticed that this will never be reached, can we delete these two lines?

xevisalle (approved):
LGTM, just some nits

src/fft/polynomial.rs
moCello: I had to check myself, but we are indeed reaching these lines. The if-else returns a result which is stored in the mutable `result` variable, which is then being truncated at the end.

xevisalle: Ohh true! I missed that part!

xevisalle (approved):

PR #807, opened 2023-12-21
Improve API for circuit compression

### LINKED ISSUE #804: Improve API for circuit (de-)compression
#### Summary
When decompressing a circuit, we immediately compile the circuit and return the `prover` and `verifier`, and because of this we require the `label` and `publicparameter` at decompression time which is counter intuitive.
At the same time we compress the circuit as part of the compile module which is also counter intuitive.
We can improve the API as follows:
Current API:
```rust
let compressed = Compiler::compress::<DummyCircuit>().unwrap();

let (decompressed_prover, decompressed_verifier) =
    Compiler::decompress(&pp, label, &compressed).unwrap();
```
Proposed new API:
```rust
let compressed = DummyCircuit::compress().unwrap();

let (decompressed_prover, decompressed_verifier) =
    Compiler::compile_with_compressed(&pp, label, &compressed).unwrap();
```

moCello:
Resolves #804

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Move compress module under composer
   - Move constraint_system module under composer
   - Move permutation module under composer
+- Change API for circuit (de-)compression [#804]
 
 ### Removed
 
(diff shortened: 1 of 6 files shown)


HDauven (approved):
LGTM

Issue #805, opened 2023-12-21
Restructure modules


moCello:
#### Summary
Restructure modules:
- move compiler to root
- move prover and verifier under compiler
- move compress under composer
- move constraint_system under composer
- move permutation under composer

Issue #802, opened 2023-12-20
Change Composer into a struct


moCello:
#### Summary
Currently we have the `Composer` trait that encodes the functionalities for composing a circuit.
Since we only ever implement this trait onto the `Builder`, we an morph the trait and struct into a single `Composer` struct. This comes with the added benefit that we can hide some of the trait functionalities from the public API that are now exposed but marked as deprecated (e.g. `uninitialized`, `append_witness_internal`, `append_custom_gate_internal`).

Issue #797, opened 2023-12-14
Inconsistent gate ordering


moCello:
#### Summary
Within the proof system, variables are not always listed in the same order. One place where this could lead to a problem is in the arithmetic proving and verifying keys. The arithmetic prover key has “q_c before q_4”, unlike the arithmetic verifier key struct which stores “q_4 before q_c”, and strangely the verifier key serialization stores “q_c before q_4”.
The arithmetic verifier key does correctly swap q_c and q_4 in from_bytes, so there are no bugs present currently. However, swapping the order throughout the codebase is very unexpected and may lead to bugs in the future.

Recommendation: pick one ordering and stick to it throughout the library. In particular, please serialize things in the same order they are stored in the struct.

#### Relevant Context
Finding of the audit.

PR #809, opened 2024-01-03
Release 0.19.0

moCello:
### Fixed

- Fix inconsistency in gate ordering of arithmetic verifier key [#797]
- Fix leading coefficients might be zero [#796]
- Fix tests when default features are turned off by placing them behind the `alloc` feature

### Changed

- Improve InvalidCircuitSize error [#792]
- Hide all modules except 'prelude' [#782]
- Turn `Composer` trait into a struct [#802]
- Rename `Arithmetization` to `Gate` [#802]
- Change internal module structure [#805]:
  - Move compiler module to root
  - Move prover and verifier modules under compiler
  - Move compress module under composer
  - Move constraint_system module under composer
  - Move permutation module under composer
- Change API for circuit (de-)compression [#804]

### Removed

- Remove `Builder` struct with introduction of `Composer` struct [#802]
- Remove example from README in favor of an actual example in the example directory that is behind the `alloc` feature [#346]

### Added

- Add example for circuit creation [#346]

[0.19.0]: https://github.com/dusk-network/plonk/compare/v0.18.0...v0.19.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.19.0] - 2024-01-03
+
 ### Fixed
 
 - Fix inconsistency in gate ordering of arithmetic verifier key [#797]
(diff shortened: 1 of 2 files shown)


herr-seppia (approved):
LGTM

PR #811, opened 2024-01-19
Updated rust doc link

sqltrigger:
Came across an empty link while reading the docs : https://docs.rs/plonk/

Updated it to current rust docs link : https://docs.rs/dusk-plonk/latest/dusk_plonk/

Hope I could be of any help :)


xevisalle:
Closing since it does not longer apply.

PR #830, opened 2024-06-27
Make code to match the last Plonk paper

xevisalle:
This PR makes the protocol to match the last version of the Plonk paper. Furthermore, it includes a document specifying our implementation of Plonk, which deviates from the original one to include custom gates.

## Main changes

The main changes are:

- Make round 4 and 5 of the prover to match the paper.
- Update the computation of the linearization polynomial `r`.
  - The prover now adds the evaluation of the public inputs polynomial at `z_challenge` to `r`.
  - The prover now subtracts the factor multiplying the evaluation of the vanishing polynomial at `z_challenge` by the quotient polynomial from `r`. As such, the quotient polynomial is not needed to compute `W_z(X)`.
  - The verifier now computes `r_0_eval`. As such, the verifier no longer needs to receive `r_eval`.
  - The verifier now subtracts the factor multiplying the evaluation of the vanishing polynomial at `z_challenge` by the quotient polynomial commitment from `[D]_1`. As such, the verifier doesn't need to compute `t_eval`, as it is not needed to compute `[E]_1`.
- The verifier adds `u_challenge` when computing the polynomial commitment `[D]_1`.
 

## Minor changes

Some minor changes have been addressed:

- Rename some variables to match the paper, making it easier to assess the correctness of the protocol.
- Refactor the proof verification to improve the efficiency and to make the code more clear.
- Other minor fixes.

## Collateral effects

The above changes have led to the following collateral effects:

- Proof verification is now 40% faster (~ from `6ms` to `3.7ms` on `Apple M1`).
- Proof size has been reduced by `1 BlsScalar`.


marta-belles (approved):
LGTM

Issue #842, opened 2025-01-31
Add component to subtract points in-circuit


xevisalle:
#### Summary
Our composer doesn't have any component that allows to subtract points in circuit. This feature would be beneficial, for instance, to reduce the number of constraints when decrypting ElGamal ciphertexts in-circuit.

Such a feature can be implemented by simply negating the 'x' coordinate of a WitnessPoint (being sure that this is enforced in-circuit), and then performing the standard point addition.

sam-ejma081:
Hello, we tried to solve the issue.

This is what we did:

Implemented a new `component_sub_point` method in the Composer struct to perform point subtraction. This method negates the x-coordinate of the second point and then uses the existing `component_add_point` method to perform the subtraction. Also added a test case for the new functionality.

You can review changes in this commit: https://github.com/sam-ejma081/dusk-network-plonk-842/commit/ad6d18a4226be3b809d0361cc224f5a14cf421fa.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

PR #843, opened 2025-01-31
Add component_sub_point to composer

xevisalle:
Resolves: #842

### DIFF
--- a/src/composer.rs
+++ b/src/composer.rs
@@ -749,6 +749,23 @@ impl Composer {
         );
     }
 
+    /// Subtracts a curve point from another by consuming 3 gates.
+    pub fn component_sub_point(
+        &mut self,
+        a: WitnessPoint,
+        b: WitnessPoint,
+    ) -> WitnessPoint {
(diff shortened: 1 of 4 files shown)


src/composer.rs
```diff
@@ -749,6 +749,23 @@ impl Composer {
...
+        &mut self,
+        a: WitnessPoint,
+        b: WitnessPoint,
+    ) -> WitnessPoint {
+        // We negate the 'x' coordinate of the point 'b', so that neg_b = (-b.x,
+        // b.y)
+        let constraint = Constraint::new().left(-BlsScalar::one()).a(*b.x());
+        let neg_b_x = self.gate_mul(constraint);
+
+        let neg_b = WitnessPoint::new(neg_b_x, *b.y());
```
moCello, line 763: I would even go as far as to add a separate `component_neg_point` as well so that we can use this in more places.

moCello (commented):
General LGTM, although while we're at it, I would also add a `component_neg_point`.

marta-belles (approved):
LGTM

PR #865, opened 2026-06-12
Release dusk-plonk 0.22.1

HDauven:
## Summary

- merge the Plonk 0.22 maintenance history up to private commit `f63cfb0c4a5e5f485361c83ae15584257d3339ab`
- keep the public `v0.22.0` release commit in history
- release the imported maintenance changes as `dusk-plonk 0.22.1`

## Notes

This keeps the Rusk 1.7 dependency line on `0.22.x` instead of moving it to the broader `0.23.0` dependency update.

## Validation

- `cargo package --allow-dirty --no-verify`
- `make test`
- `make clippy`
- `make no-std`
- `cargo publish --dry-run`

### DIFF
--- a/src/proof_system/proof.rs
+++ b/src/proof_system/proof.rs
@@ -375,9 +375,9 @@ pub(crate) mod alloc {
 
             // Evaluations to compute [E]_1
             //
-            // IMPORTANT: Ordering must match the prover's batched opening at `z`
-            // (`CommitKey::compute_aggregate_witness([...])`) and the verifier's
-            // commitment list below.
+            // IMPORTANT: Ordering must match the prover's batched opening at
+            // `z` (`CommitKey::compute_aggregate_witness([...])`)
+            // and the verifier's commitment list below.
             let E_evals = vec![
                 // Unshifted openings at z
                 self.evaluations.a_eval,
@@ -1062,11 +1062,11 @@ mod soundness_tests {
     ///
     /// The proof uses:
     /// - Honest wire polynomials (circuit is satisfied at gate level)
-    /// - A RANDOM permutation polynomial z (breaking the permutation
-    ///   argument — gates are not properly wired together)
+    /// - A RANDOM permutation polynomial z (breaking the permutation argument —
+    ///   gates are not properly wired together)
     /// - RANDOM quotient polynomials (not the actual quotient)
-    /// - A FORGED `q_arith_eval` computed to balance the verification
-    ///   equation after observing `z_challenge`
+    /// - A FORGED `q_arith_eval` computed to balance the verification equation
+    ///   after observing `z_challenge`
     ///
     /// This proof is invalid (the permutation argument does not hold)
     /// but exploits the fact that selector evaluations are not bound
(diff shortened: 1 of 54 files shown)


src/proof_system/proof.rs
```diff
@@ -947,15 +907,59 @@ pub(crate) mod alloc {
 
         result * numerator
     }
+
+    pub(crate) fn compute_barycentric_eval_sparse(
```
moCello, line 911: **NOTE:** New public-input evaluator `compute_barycentric_eval_sparse` has no direct unit test pinning it against the dense path it replaces; the verifier's soundness rests on the two producing the same value for every (index, value) set.

**Context:** It is exercised only indirectly, on the V2/V3 path: prove→verify round-trips plus soundness tests that forge with the dense `compute_barycentric_eval` (proof.rs:857) and verify through this sparse path. Untested: a direct sparse-vs-dense assertion over non-contiguous index sets (where the two could diverge), the empty-indexes return (proof.rs:917), and the all-zero-evaluations return (proof.rs:932).

**Fix:** Add a unit test asserting `compute_barycentric_eval_sparse(indexes, sparse_vals, z, domain)` equals `compute_barycentric_eval(dense_vals, z, domain)` across a few randomized non-contiguous index sets, plus the empty-indexes and all-zero-evaluations cases.

src/proof_system/proof.rs
```diff
@@ -517,6 +497,7 @@ pub(crate) mod alloc {
             verifier_key: &VerifierKey,
             transcript: &mut Transcript,
             opening_key: &OpeningKey,
+            public_input_indexes: &[usize],
```
moCello, line 500: **NOTE:** Rewritten `verify_legacy` body (MSM regrouping + sparse PI eval) has no test coverage — no test reaches the V1 verify path, and V1 proving is disabled so the suite cannot generate a proof to round-trip it.

**Context:** `verify_legacy` is reached only via `verify_with_version(.., PlonkVersion::V1)` (compiler/verifier.rs:225), which no test calls. `prove_with_version(.., V1)` returns `UnsupportedProvingVersion` (compiler/prover.rs:252), so the suite cannot produce a V1 proof to exercise the rewrite even under `--all-features`. This is the bulk of the codecov-flagged 18 lines. The degree-7 math mirrors the V2/V3 (degree-11) rewrite and is identical by inspection, but this compat path — the one that verifies historical proofs — ships with no regression safety net.

**Fix:** Add a regression test verifying a stored historical V1 proof through `verify_with_version(.., PlonkVersion::V1)`; or, if legacy verification is being retired, document that the path is intentionally untested and why.

moCello (approved):
APPROVE — Release looks sound. Overflow guards in both deserializers are correct and tested; the verifier MSM/PI-eval refactor is algebraically identical to the code it replaces, for both current and legacy batching, so no soundness regression by inspection. CHANGELOG, version bump, and commit history are clean. Two non-blocking coverage asks: (1) add a direct unit test pinning `compute_barycentric_eval_sparse` against the retained dense evaluator, covering its empty and all-zero branches; (2) the rewritten `verify_legacy` path has no test coverage at all — no test verifies a V1 proof, and V1 proving is disabled so the suite cannot generate one — so add a stored-proof regression test or document the gap.

PR #868, opened 2026-07-20
Bind logic gadget output to its inputs

### LINKED ISSUE #867: Bind recomposed operands to inputs in append_logic_component
`append_logic_component` constrains the per-quad accumulators among themselves but doesn't tie the final recomposed operands back to the declared inputs `a`/`b`. This adds a closing binding (a range-bounded recomposition) so the gate fully enforces its intended relation.

Two effects to note:
- The gate's constraints change → **new verifying key** (downstream proofs/keys need regenerating).
- The maximum width becomes explicit — `BIT_PAIRS <= 127` (254 bits), erroring above that rather than the previous silent cap at 256.

Ships as a breaking release.

moCello:
The logic gadget reconstructed its operands from the bits of `a` and `b` without wiring the input witnesses into any gate, leaving the output decoupled from the inputs. A prover could satisfy the identical gate layout — and so the same verifier key — with arbitrary accumulators, for instance proving `0 & 0 = 0` while declaring `a = 3, b = 3`.

Close the gadget with a binding that ties each reconstructed accumulator back to the low bits of its input. The binding is a range-bounded canonical truncation rather than a plain equality, so inputs wider than the op width are still accepted; a plain `assert_equal` would reject honest inputs.

The op width is capped at 254 bits, the widest even width for which the truncation binding stays canonical against a 255-bit `BlsScalar`. `BIT_PAIRS > 127` is now a compile-time error rather than a silent cap.

Extract the shared pieces this needs: `range_check`/`range_check_even` as the single base-4 range layout, `component_truncate` as the canonical truncation primitive, and `component_range_bits` as a bit-counted range check, deprecating the bit-pair-counted `component_range`.

This changes the verifier key of any circuit using the logic gadget.

### DIFF
--- a/.github/workflows/dusk_ci.yml
+++ b/.github/workflows/dusk_ci.yml
@@ -69,7 +69,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: Swatinem/rust-cache@v2
       - run: sudo apt install -y kcov
-      - run: RUSTFLAGS='-Cdebuginfo=2 -Cinline-threshold=0 -Clink-dead-code' cargo test --release --no-run --all-features
+      - run: RUSTFLAGS='-Cdebuginfo=2 -Cllvm-args=--inline-threshold=0 -Clink-dead-code' cargo test --release --no-run --all-features
       - name: Collect coverage
         run: >
           find target/release/deps -type f -executable ! -name "*.*" |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `Composer::component_truncate<N>` to extract the low `N` bits of a witness [#867]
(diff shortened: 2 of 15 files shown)


xevisalle (approved):
lgtm

HDauven (approved):
The constraint construction and canonicality maths look sound to me. I did find one issue in the regression tests that I think should be fixed, but not necessarily block this PR.

The canonicality tests can remain green if the production `assert_canonical_truncation` call is removed. The forged branches manually reproduce the guard and accept any proving error, so a resulting `InvalidCircuitSize` is treated as successful rejection. Additionally, the truncation alias test compiles with the default input `-1` but proves the alias with input `1`, changing the constant selector and causing an independent rejection.

I confirmed this through mutation testing, removing the production guard did not fail the canonicality tests.

Could we compile and prove using the same fixed input, assert that the honest and forged gate layouts are identical and explicitly reject `InvalidCircuitSize`/unrelated errors?

Minor documentation point, the performance comparison is reversed after binding both logic inputs. I measured `component_truncate::<2>` at 88 constraints versus 172 for XOR-with-zero, and 88 versus 234 at 250 bits. Truncation is cheaper at narrow widths too.

PR #869, opened 2026-07-22
Add a prime-order subgroup check for witness points

### LINKED ISSUE #870: Add a prime-order subgroup check for witness points
The composer's point gadgets operate on `WitnessPoint`s that are assumed to be valid group elements, but there is currently no in-circuit way to constrain a prover-supplied point into the prime-order subgroup. This adds a composer method that enforces prime-order membership for a witness point, so a caller can validate an untrusted point once at its entry boundary.

**Scope update:** the fix lands as more than the single composer method. Alongside `Composer::assert_torsion_free_point`, a `TorsionFreeWitnessPoint` type encodes subgroup membership in the type system, with **breaking API changes**:

- `component_add_point`, `component_sub_point`, `component_mul_point`, `component_neg_point` and `component_select_identity` take and return `TorsionFreeWitnessPoint`;
- `Composer::IDENTITY` becomes a `TorsionFreeWitnessPoint`;
- `append_constant_point` validates its constant natively and returns `Result<TorsionFreeWitnessPoint, Error>` (new `Error::JubJubPointNotTorsionFree` variant);
- `assert_torsion_free_point` returns the typed point;
- `component_mul_generator` returns `Result<TorsionFreeWitnessPoint, Error>` — its generator base is validated to exact prime order, so the multiples carry the type directly;
- `component_select_identity` constrains its `bit` to be boolean itself (one additional gate), so the typed return cannot be subverted by a non-boolean assignment. This changes the circuit shape for direct callers, which need new verifier keys; `component_mul_point` is unaffected.

Downstream circuits building on these gadgets need adjusting for the release: membership is established at each point's entry (in-circuit via `assert_torsion_free_point` for prover-controlled points, `append_constant_point` for constants, or `TorsionFreeWitnessPoint::new_unchecked` for points validated off-circuit by the accepting verifier's protocol) and is preserved by the group operations from there.

moCello:
Resolves #870

### DIFF
--- a/src/composer.rs
+++ b/src/composer.rs
@@ -584,9 +584,13 @@ impl Composer {
         self.assert_canonical_truncation(high, low, num_bits);
     }
 
-    /// Evaluate `jubjub · generator` as a [`WitnessPoint`].
+    /// Evaluate `jubjub · generator` as a [`TorsionFreeWitnessPoint`].
     ///
-    /// `generator` is appended to the circuit description as a constant.
+    /// `generator` is appended to the circuit description as a constant. It
+    /// is validated to be an on-curve point of exact prime order, so every
+    /// multiple of it lies in the prime-order subgroup and the result carries
+    /// the [`TorsionFreeWitnessPoint`] membership by construction; see that
+    /// type for the boundary rule.
     ///
     /// The circuit constrains `jubjub` to the canonical Jubjub scalar interval
     /// `[0, r)`, where `r` is the prime-order subgroup modulus. It also bounds
@@ -626,18 +630,11 @@ impl Composer {
     /// generation if `jubjub` is not a canonical Jubjub scalar. Soundness does
     /// not rely on that host-side check: the same canonicality and signed-digit
     /// bounds are enforced by circuit constraints.
-    ///
-    /// The returned point is untyped: the `generator` base is caller-supplied
-    /// and not validated here, so this routine cannot vouch for subgroup
-    /// membership of its multiples. A caller passing a generator known to lie
-    /// in the prime-order subgroup can wrap the result with
-    /// [`TorsionFreeWitnessPoint::new_unchecked`]; see
-    /// [`TorsionFreeWitnessPoint`] for the boundary rule.
     pub fn component_mul_generator<P: Into<JubJubExtended>>(
         &mut self,
         jubjub: Witness,
         generator: P,
-    ) -> Result<WitnessPoint, Error> {
+    ) -> Result<TorsionFreeWitnessPoint, Error> {
         let generator = generator.into();
 
         if !bool::from(generator.is_on_curve() & generator.is_prime_order()) {
@@ -662,7 +659,10 @@ impl Composer {
             "the wnaf_entries array is expected to be 256 elements long"
         );
 
+        // The generator passed the exact prime-order validation above, so
+        // its multiples cannot leave the prime-order subgroup.
         self.append_fixed_base_signed_digits(jubjub, generator, &wnaf_entries)
+            .map(TorsionFreeWitnessPoint::new_unchecked)
     }
 
     /// Constrain a fixed-base multiplication for a supplied signed-digit
--- a/src/composer.rs
+++ b/src/composer.rs
@@ -1074,6 +1074,11 @@ impl Composer {
     }
 
     /// Subtracts a curve point from another by consuming 3 gates.
+    ///
+    /// As with [`Self::component_add_point`], point validity (on-curve and,
+    /// where it matters, prime-order subgroup membership) is a **caller
+    /// obligation** discharged at the point's entry boundary — this gate
+    /// assumes valid group elements and does not re-check them.
     pub fn component_sub_point(
         &mut self,
         a: WitnessPoint,
@@ -1087,6 +1092,19 @@ impl Composer {
     }
 
     /// Adds two curve points by consuming 2 gates.
+    ///
+    /// # Point validity is a caller obligation
+    ///
+    /// This gate constrains only the twisted-Edwards addition identity; it does
+    /// **not** check that `a`/`b` lie on the curve or in the prime-order
+    /// subgroup. Point arithmetic is closed over the group, so validity is
+    /// intended to be enforced **once, at the boundary where a point enters the
+    /// circuit** — not re-checked at every operation. Callers must ensure each
+    /// [`WitnessPoint`] reaching this gate is a valid group element: an output
+    /// of [`Self::component_mul_generator`], a point bound to a
+    /// decoded/committed source, or one explicitly constrained on entry. An
+    /// off-curve or low-order input yields a well-formed proof over the wrong
+    /// group element — a soundness break in the *consumer circuit*, not here.
     pub fn component_add_point(
         &mut self,
         a: WitnessPoint,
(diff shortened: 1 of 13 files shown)


xevisalle (changes requested):
The component itself lgtm, all the constraints seem to be in place. Only issue I found is documentation-wise. Here's a finding and a suggestion:

```
The new documentation incorrectly suggests that binding/decoding alone satisfies the point-validity obligation at src/composer.rs:1030, repeated at src/composer.rs:1170.

JubJubAffine decoding guarantees an on-curve point, not prime-order subgroup membership; a commitment only binds a value. Likewise, component_mul_generator accepts an arbitrary fixed base and does not validate its subgroup. A caller following this guidance could skip the new check and remain vulnerable to low-order inputs.

Suggested wording should require:

  - a generator already known to be prime-order;
  - decoding into a subgroup-enforcing type or an explicit external subgroup check; or
  - assert_torsion_free_point inside the circuit.
 ```

Maybe we could also explicitly state that in the scenario where the input points are public, this check can be performed off-circuit.

**Review by HDauven — CHANGES_REQUESTED** [review]
The `assert_torsion_free_point` construction itself looks sound to me. `Q` is fully constrained to the curve equation, the three doublings bind all intermediate coordinates and the final equality enforces `point == [8]Q`. On Jubjub's cofactor-8 group this admits exactly the torsion-free subgroup, including the identity.

The soundness tests are also well designed and provide stronger coverage than what we had before. They preserve the production gate layout and cover low-order components, off-curve inputs, attacker-selected `Q` values, forged curve-equation intermediates and forged doubling outputs.

I am requesting changes because the caller guidance makes the point-validity invariant harder to follow than necessary.

I think the documentation should state one clear boundary rule: every point entering the circuit must already have established subgroup membership. Public or constant points may be checked off-circuit as part of the verifier's protocol, while private or prover-controlled points must use `assert_torsion_free_point`. Once that has been established, subsequent group operations preserve the invariant.

Decoding, commitment binding or passing through `component_mul_generator` should not be listed as sufficient on their own, since none of those inherently establishes subgroup membership.

Longer term, the clean API solution would be a distinct `TorsionFreeWitnessPoint` type, so the invariant is encoded in the type system rather than repeated in documentation. Might be worthy of a follow-up even.

**Comment by moCello** [comment]
@HDauven @xevisalle — agreed on the documentation finding: decoding, commitment binding and `component_mul_generator` don't establish subgroup membership on their own, and the guidance should state one boundary rule instead of listing shortcuts.

Rather than fixing this in prose only, I want to take up the `TorsionFreeWitnessPoint` suggestion in this PR. We're already in a breaking window, and a doc-only rule written now would be rewritten as type docs the moment the newtype lands.

Planned shape — input welcome if you disagree with the shape:

- `TorsionFreeWitnessPoint`: Copy wrapper around `WitnessPoint`, inner field private. Torsion-free includes the identity; consumers needing non-identity add that constraint separately (documented on the type).
- Constructors carry the boundary rule:
  - `assert_torsion_free_point(WitnessPoint) -> TorsionFreeWitnessPoint` — the in-circuit path for prover-controlled points (the check already in this PR).
  - `TorsionFreeWitnessPoint::new_unchecked(WitnessPoint)` — for points whose membership is established off-circuit (public inputs and constants checked as part of the verifier's protocol). The caller obligation is documented here, once.
  - `append_constant_point` checks is_torsion_free natively at circuit build and returns the typed point.
  - `component_mul_generator` returns the typed point once the generator validation from #873 is in — its base is then guaranteed prime-order. Until that lands it keeps returning `WitnessPoint`.
- `component_add_point`,  `component_sub_point` and `component_mul_point` take and return `TorsionFreeWitnessPoint`, so group closure is encoded in the signatures — once in the subgroup, a point can't leave it.
- CHANGELOG gets a breaking-API entry.

If the shape looks right to you I'll push the type change and the reworded docs together.

**Comment by HDauven** [comment]
> If the shape looks right to you I'll push the type change and the reworded docs together.

The shape LGTM. I would much prefer this over just leaving a documentation obligation behind.

Few details to account for:
1. I would treat `new_unchecked` as a very explicit escape hatch rather than the normal public input path. `Verifier::verify` only receives flat scalar inputs, so PLONK itself cannot know that an external subgroup check actually happened. I think keeping it `pub(crate)` would be preferable unless  downstream custom gadgets need it. (Phoenix?) If we do keep it public, the docs should make clear that the accepting verifier must enforce the check, not only the honest prover.
2. `append_constant_point` should check both `is_on_curve` and `is_torsion_free` and return an error rather than panic. `is_torsion_free` alone does not validate unchecked raw coordinates.
3. As far as I can see we will still need a private raw addition primitive. The `Q` used inside `assert_torsion_free_point` only needs to be on curve and may contain a torsion part before `[8]Q`, so it should not be wrapped as `TorsionFreeWithnessPoint`, unless I'm mistaken.
4. Ideall this should stack on #873 so `component_mul_generator` returns the typed point from the start, rather than temporarily going through `new_unchecked`.
5. If possible, I would also carry the type through identity and negation.

With those points accounted for, +1 from me. In my opinion this is a better end state than repeating the boundary rule throughout the API. Let me know what you think.

**Comment by xevisalle** [comment]
the plan lgtm, I'm finishing addressing the leftovers PR comments so we can merge asap

File: src/composer.rs [hunk 1]
```diff
@@ -1026,44 +1198,54 @@ impl Composer {
...
         bit: Witness,
-        a: WitnessPoint,
-    ) -> WitnessPoint {
+        a: TorsionFreeWitnessPoint,
+    ) -> TorsionFreeWitnessPoint {
         let x = self.component_select_zero(bit, *a.x());
         let y = self.component_select_one(bit, *a.y());
 
-        WitnessPoint::new(x, y)
+        TorsionFreeWitnessPoint::new_unchecked(WitnessPoint::new(x, y))
```
HDauven: `component_select_identity` still relies on the caller constraining `bit` to be boolean, but returns a `TorsionFreeWitnessPoint`.

For a non-boolean `bit = t`, this produces `(t·x, 1 - t + t·y)`, which is generally not on-curve. The existing invalid-bit test uses identity, which hides this because identity maps back to identity for any `t`.

Can we constrain `bit` here, use a boolean witness type or keep the typed helper private and only use it from `component_mul_point`, where the bits are already constrained?

HDauven (changes requested):
I really like where this PR ended up. The tests are much stronger and encoding the invariant in the API is much better than leaving it to every caller to do correctly.

There is still one caller invariant inside the typed API in `component_select_identity` where the typed invariant depends on an unenforced caller obligation, see comment.

src/composer.rs
moCello: Constrained it here. `component_select_identity` now boolean-constrains `bit` itself (+1 gate), so the typed return cannot be subverted by a non-boolean assignment. Added a rejection test for exactly the `(t·x, 1 - t + t·y)` off-curve case you described — the old invalid-bit test leaned on identity, which as you noted hid it, so that expectation is flipped.

`component_mul_point` is unchanged: it calls a private `select_identity_gates` seam whose bits are already constrained by `component_decomposition`, so no extra gate there and its layout is pinned by a gate-digest golden. Commit 93776c6.

HDauven (approved):
LGTM!

xevisalle (approved):
lgtm

PR #871, opened 2026-07-27
Fix generator multiplication

xevisalle:
Closes: https://github.com/dusk-network/plonk-private/issues/30

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Cap logic gadget width at 254 bits; `BIT_PAIRS > 127` is now a compile-time error [#867]
 - Change the verifier key of circuits using the logic gadget [#867]
 - Increase the gate count of `append_logic_and`/`append_logic_xor` to bind their inputs [#867]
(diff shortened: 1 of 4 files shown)


HDauven (approved):
LGTM

I verified the fix against the fixed-base gate semantics:
* `d_{i+1} - 2d_i` is constrained to `{-1, 0, 1}`. Requiring the accumulator after the first three most-significant rounds to equal zero uniquely forces those three digits to zero.
* The remaining 253 signed digits have absolute value below `2^253`. Combined with the in-circuit canonical input bound `0 <= scalar < r < 2^252`, the endpoint difference has absolute value below `2^254 < q`. The closing BLS-field equality therefore cannot hide a `±q` wrap.
* The second 252-bit range check correctly enforces `scalar <= r - 1` by bounding `(r - 1) - scalar`.
* The shifted-wire anchor remains immediately after the 256th fixed-base row.
* The adversarial regression tests preserve the production gate layout and cover `q`, `q + 1`, `1 - q`, `r`, `r - 1`, and the high carry boundaries.

Operationally, every affected circuit-specific proving key, verifier key, and cached compressed circuit description must be regenerated before deployment, so we need to bundle this PR with #869.

PR #872, opened 2026-07-27
Fix append bug

xevisalle:
Closes: https://github.com/dusk-network/plonk-private/issues/35


HDauven (approved):
LGTM

Fixes the issue as intended. `append_evaluated_output` now allocates the computed output, wires it into the constraint and appends one active arithmetic row. `gate_add` and `gate_mul` delegate to it without duplicating the row, so their existing layouts remain unchanged.

The regression tests verify the honest case, reject the shape-compatible `2 + 3 = 9` substitution, compare the complete gate layout and cover zero, one, minus one and arbitrary nonzero output selectors.

One minor documentation suggestion: direct callers that previously appended their own gate should remove that manual insertion, in addition to regenerating their circuit keys and compressed descriptions.

PR #873, opened 2026-07-29
Tackle leftovers

### LINKED ISSUE #832: Look into leftovers
#### Summary

There are some leftovers in the code that need to be addressed, i.e. FIXMEs, TODOs, etc. In particular:

- https://github.com/dusk-network/plonk/blob/499a70f5da81ade4dbd226f00413a7a17e96f857/src/composer.rs#L507
- This list in WIP

xevisalle:
Closes: https://github.com/dusk-network/plonk/issues/832

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `Composer::component_truncate<N>` to extract the low `N` bits of a witness [#867]
 - Add `Composer::component_range_bits<BITS>` range check counting bits directly [#867]
+- Add the public `Error::JubJubGeneratorNotPrimeOrder` variant [#867]
 
 ### Changed
 
@@ -22,6 +23,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   for direct callers, which must regenerate circuit-specific proving and
   verifier keys and cached compressed circuit descriptions. The `gate_add` and
   `gate_mul` wrapper layouts remain one row and are unchanged.
+- Make `Composer::component_mul_generator` reject generators that are not
+  on-curve points of exact prime order. Callers using custom generators should
+  re-audit them; rejected generators now return
+  `Error::JubJubGeneratorNotPrimeOrder` [#867]
 - Constrain `Composer::component_mul_generator` inputs to canonical Jubjub
   scalars and its signed-digit recurrence to 253 effective digits. This changes
   the circuit shape: applications using the component must regenerate their
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `Composer::component_truncate<N>` to extract the low `N` bits of a witness [#867]
 - Add `Composer::component_range_bits<BITS>` range check counting bits directly [#867]
-- Add the public `Error::JubJubGeneratorNotPrimeOrder` variant [#867]
+- Add the public `Error::JubJubGeneratorNotPrimeOrder` variant [#832]
 
 ### Changed
 
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Make `Composer::component_mul_generator` reject generators that are not
   on-curve points of exact prime order. Callers using custom generators should
   re-audit them; rejected generators now return
-  `Error::JubJubGeneratorNotPrimeOrder` [#867]
+  `Error::JubJubGeneratorNotPrimeOrder` [#832]
 - Constrain `Composer::component_mul_generator` inputs to canonical Jubjub
   scalars and its signed-digit recurrence to 253 effective digits. This changes
   the circuit shape: applications using the component must regenerate their
@@ -716,6 +716,7 @@ is necessary since `rkyv/validation` was required as a bound.
 [#859]: https://github.com/dusk-network/plonk/issues/859
 [#842]: https://github.com/dusk-network/plonk/issues/842
 [#834]: https://github.com/dusk-network/plonk/issues/834
+[#832]: https://github.com/dusk-network/plonk/issues/832
 [#831]: https://github.com/dusk-network/plonk/issues/831
 [#819]: https://github.com/dusk-network/plonk/issues/819
 [#818]: https://github.com/dusk-network/plonk/issues/818
--- a/src/commitment_scheme/kzg10/key.rs
+++ b/src/commitment_scheme/kzg10/key.rs
@@ -278,6 +278,28 @@ pub struct OpeningKey {
     pub(crate) prepared_x_h: G2Prepared,
 }
 
+fn batch_challenge(
+    transcript: &mut Transcript,
+    points: &[BlsScalar],
+    proofs: &[Proof],
+) -> BlsScalar {
+    transcript.append_message(b"dom-sep", b"kzg10-batch-check-v1");
+    transcript.append_u64(b"batch-len", proofs.len() as u64);
+    for (point, proof) in points.iter().zip(proofs) {
+        transcript.append_scalar(b"batch-point", point);
+        transcript.append_commitment(
+            b"batch-polynomial-commitment",
+            &proof.commitment_to_polynomial,
+        );
+        transcript.append_scalar(b"batch-evaluation", &proof.evaluated_point);
+        transcript.append_commitment(
+            b"batch-witness-commitment",
+            &proof.commitment_to_witness,
+        );
+    }
+    transcript.challenge_scalar(b"batch-challenge")
+}
+
 impl Serializable<{ G1Affine::SIZE + G2Affine::SIZE * 2 }> for OpeningKey {
     type Error = dusk_bytes::Error;
 
@@ -317,8 +339,16 @@ impl OpeningKey {
         }
     }
 
-    /// Checks whether a batch of polynomials evaluated at different points,
-    /// returned their specified value.
+    /// Checks whether a batch of polynomials evaluated at different points
+    /// returned their specified values.
+    ///
+    /// The caller-provided transcript supplies any surrounding protocol
+    /// context. This method then binds the batch length and every opening
+    /// point, polynomial commitment, evaluation, and witness commitment before
+    /// deriving the random linear-combination challenge. The challenge
+    /// therefore cannot be known before the complete batch is fixed, and no
+    /// additional verifier randomness is required.
+    // Retained for planned production batch verification.
     #[allow(dead_code)]
     pub(crate) fn batch_check(
         &self,
@@ -326,10 +356,14 @@ impl OpeningKey {
         proofs: &[Proof],
         transcript: &mut Transcript,
     ) -> Result<(), Error> {
+        if proofs.is_empty() || points.len() != proofs.len() {
+            return Err(Error::ProofVerificationError);
+        }
+
         let mut total_c = G1Projective::identity();
         let mut total_w = G1Projective::identity();
 
-        let u_challenge = transcript.challenge_scalar(b"batch"); // XXX: Verifier can add their own randomness at this point
+        let u_challenge = batch_challenge(transcript, points, proofs);
         let powers = util::powers_of(&u_challenge, proofs.len() - 1);
         // Instead of multiplying g and gamma_g in each turn, we simply
         // accumulate their coefficients and perform a final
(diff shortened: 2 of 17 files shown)


src/proof_system/widget.rs
HDauven: This still leaves an unchecked deserialization path.

`evaluations_size` comes from the serialized prover key and is passed straight into `split_at`, so malformed input can panic here instead of returning an error.

There is also no check in `Evaluations::from_slice` that the number of evaluations actually matches the serialized domain size. Arithmetic only compares the domains and then zips the vectors, so mismatched lengths are silently accepted here.

Can we use checked slicing here, validate `evals.len() == domain.size()`, and add tests for malformed lengths and domains?

I would also prefer the prover key parser to verify that these evaluation sets use the expected `8n` domain.

HDauven (changes requested):
The rest looks good to me. The KZG batch challenge now binds the complete batch, the fixed-base generator check correctly excludes identity and non-prime-order points. I did not find an issue in the linearization refactor or debugger gate equations.

I am requesting changes for the unchecked proving key deserialization path though. This is not a soundness break but malformed input can currently panic instead of returning an error, and the serialized evaluation/domain cardinality invariant is not enforced.

src/proof_system/widget.rs
```diff
@@ -441,10 +441,6 @@ pub(crate) mod alloc {
             let mut buffer = bytes;
             let n = u64::from_reader(&mut buffer)? as usize;
             let evaluations_size = u64::from_reader(&mut buffer)? as usize;
```
moCello: **WARNING:** `evaluations_size`, read from the serialized blob, flows unchecked into `buf.split_at(evaluations_size)` (widget.rs:463) — a malformed prover key panics instead of returning an error.

**Context:** The reachable entry is public: `Prover::try_from_bytes` (prover.rs:151) validates only the *outer* segment lengths (`required_len`, prover.rs:195), then passes the prover-key segment to `ProverKey::from_slice`. `evaluations_size` is an inner length never bounded against that segment, so a blob with a valid outer frame but an oversized inner `evaluations_size` reaches `split_at` and panics; `serialized_poly_len` (widget.rs:454) has the same shape. Separately, `Evaluations::from_slice` (evaluations.rs:64) never checks `evals.len() == domain.size()`, so a set whose length disagrees with its domain is accepted and only fails later inside a `zip`. This is the concern the prior change request is open on; the PR deleted the comment that documented the gap but left the code unchanged.

**Severity-rationale:** Pre-existing code the PR touches. The input is a locally-produced prover-key artifact, not adversarial network data — a verifier receives a `Proof` + `VerifierKey`, never a `ProverKey` — so this is a robustness gap on corrupted caches, not a remote attack surface, which holds it below blocker. It stays a warning rather than a note because the same file asserts a tested no-panic contract for this deserializer (`prover_try_from_bytes_rejects_*_without_panicking`, wrapping `try_from_bytes` in `catch_unwind`) that no test exercises on a malicious inner length.

**Fix:** Replace the `split_at` calls with a length-checked split returning `Error` on overrun, validate `evals.len() == domain.size()` in `Evaluations::from_slice` and the expected `8n` domain in `ProverKey::from_slice` (which knows `n`), and add tests feeding oversized/undersized lengths and mismatched domains.

src/error.rs
```diff
@@ -79,6 +79,9 @@ pub enum Error {
     /// This error occurs when a malformed JubJub scalar is decoded from a byte
     /// array.
     JubJubScalarMalformed,
+    /// This error occurs when a fixed-base generator is not a prime-order
```
moCello, line 82: **WARNING:** The one consumer-observable change in this PR — a new public `Error` variant plus `component_mul_generator` now rejecting non-prime-order generators — has no CHANGELOG entry.

**Context:** Nothing under `[Unreleased]` records `Error::JubJubGeneratorNotPrimeOrder` or the behavior change where `component_mul_generator` returns `Err` for a generator it previously accepted. `Error` is not `#[non_exhaustive]`, so the added variant is itself a potential source-breaking change for any downstream exhaustive `match`. The acceptance change is soundness-relevant: the fixed-base gadget's correctness assumes a prime-order generator, so a consumer that previously passed a non-prime-order point should re-audit, not just recompile.

**Fix:** Add an `### Added` line for the new error variant and a `### Changed` (or `### Fixed`) line for the generator validation under `[Unreleased]`, referencing the tracking issue via a `[#nnn]` link definition as the block already does.

src/debugger.rs
```diff
@@ -26,6 +27,158 @@ pub(crate) struct Debugger {
...
+    }
+
+    fn delta(value: BlsScalar) -> BlsScalar {
+        value
+            * (value - BlsScalar::one())
+            * (value - BlsScalar::from(2u64))
+            * (value - BlsScalar::from(3u64))
+    }
+
+    fn logic_identity(
```
moCello: **NOTE:** The debugger re-derives gate identities that already exist as `pub(crate)` helpers, creating two independent copies of each formula that can silently diverge.

**Context:** `Debugger::logic_identity` is a verbatim copy of `logic::proverkey::delta_xor_and`; `Debugger::delta` (debugger.rs:150) copies `logic::proverkey::delta`; and the fixed-base identities in `evaluates_to_zero` re-inline `extract_bit` and `check_bit_consistency` from `fixed_base::proverkey`. All four helpers are `pub(crate)` and reachable from the debugger. The identities match their sources today, but a later change to a widget formula updates the prover and leaves the debugger — the tool meant to say which constraint is unsatisfied — reporting against a stale identity. (`range::proverkey::delta` is identical but module-private, so sharing one `delta` there needs a visibility bump.)

**Fix:** Call the existing `pub(crate)` helpers (`delta_xor_and`, `delta`, `extract_bit`, `check_bit_consistency`) from the debugger instead of re-deriving them, so one definition drives both paths.

src/debugger.rs
```diff
@@ -26,6 +27,158 @@ pub(crate) struct Debugger {
...
+
+    fn wire_value(
+        &self,
+        constraint: &Constraint,
+        wire: WiredWitness,
+    ) -> BlsScalar {
+        self.witness_value(constraint.witness(wire))
+    }
+
+    fn shifted_wire_value(
```
moCello, line 52: **NOTE:** Rotated (`_w`) wire reads return zero past the last constraint; when the constraint count exactly fills the power-of-two domain, the real cyclic domain wraps to row 0 instead, so a last-row gate identity can be misreported.

**Context:** Range, logic, and ECC identities read `a_w`/`b_w`/`d_w` from `constraint_index + 1`. The proving domain is `constraints.next_power_of_two()` with wires zero-padded to it, so while padding exists the zero default is faithful — the real shifted row is a zero row. Only when the circuit fills the domain exactly do the last row's shifted wires come from row 0, where `unwrap_or_default()` still yields zero. A debug-only false report, bounded to that edge case.

**Fix:** Wrap the shifted index modulo the padded circuit size (`constraints().next_power_of_two()`) — padding rows still read as zero, and the exact-fill case correctly reads row 0 — or document that the last row is not evaluated.

tests/ecc.rs
```diff
@@ -432,6 +432,17 @@ fn component_mul_generator() {
...
 
+#[test]
+fn component_mul_generator_rejects_non_prime_order_generator() {
+    let mut composer = Composer::initialized();
+    let scalar = composer.append_witness(JubJubScalar::one());
+
+    let result =
+        composer.component_mul_generator(scalar, JubJubExtended::identity());
+
+    assert!(matches!(result, Err(Error::JubJubGeneratorNotPrimeOrder)));
```
moCello, line 443: **NOTE:** The new rejection test only exercises the identity point, leaving the guard's second factor (`is_torsion_free` on a non-identity torsion point) and the doc's claim that such points are rejected untested.

**Context:** The identity fails via the `!is_identity()` factor of `is_prime_order`; a genuine non-prime-order point (a small-order torsion point) exercises the `is_torsion_free()` factor instead — the branch the doc comment at composer.rs:565-568 leans on, which this test never reaches.

**Fix:** Add a case passing a known small-order / non-torsion-free Jubjub point and assert the same error.

src/commitment_scheme/kzg10/key.rs
```diff
@@ -317,19 +338,30 @@ impl OpeningKey {
...
+    /// Checks whether a batch of polynomials evaluated at different points
+    /// returned their specified values.
+    ///
+    /// The caller-provided transcript supplies any surrounding protocol
+    /// context. This method then binds the batch length and every opening
+    /// point, polynomial commitment, evaluation, and witness commitment before
+    /// deriving the random linear-combination challenge. The challenge
+    /// therefore cannot be known before the complete batch is fixed, and no
+    /// additional verifier randomness is required.
     #[allow(dead_code)]
```
moCello, line 351: **NOTE:** `batch_check` remains production-dead behind an unexplained `#[allow(dead_code)]`; its only callers are tests.

**Context:** The tracking issue asked to settle the randomness contract "before `batch_check` is exposed or used". The contract is now settled (transcript-derived challenge binding the full batch), which is the right call, but the function is still reachable only from `#[cfg(test)]`, so the suppression is load-bearing with no comment stating why it is kept.

**Fix:** Either wire `batch_check` into the verification path that needs it, or add a one-line comment on the `#[allow(dead_code)]` stating it is retained for planned batch verification.

moCello (commented):
COMMENT — the leftover cleanup is sound: every debugger gate identity matches its widget source, the generator guard correctly rejects the identity and torsion points, and the linearization refactor is a faithful field-named rename. One item gates merge, the same one the standing change request is already open on: the unchecked prover-key deserialization path. This PR deleted the comment that documented the `split_at` panic but not the panic; checked slicing, an `evals.len() == domain.size()` check, and malformed-input tests are still needed. The rest fits in that same revision without forcing another round — including the one other thing worth doing now: the sole consumer-visible change, the new public `Error::JubJubGeneratorNotPrimeOrder` and the generator rejection, has no CHANGELOG entry. The four notes are the author's call and none block: the debugger re-deriving gate helpers instead of calling them, the last-row rotation wrap, an identity-only rejection test, and a `batch_check` left dead behind an unexplained `allow`.

HDauven (approved):
LGTM!

src/compiler/prover.rs
```diff
@@ -655,6 +656,15 @@ mod tests {
...
     }
 
+    fn assert_deserialization_error_without_panic(bytes: &[u8]) {
+        let result = std::panic::catch_unwind(|| Prover::try_from_bytes(bytes));
+        assert!(result.is_ok(), "deserializer should never panic");
+        assert!(
+            result.expect("checked above").is_err(),
+            "malformed input must be rejected"
+        );
+    }
```
moCello, line 674: **NOTE:** `try_from_bytes` never checks the header against the payload: a blob whose `size` disagrees with `constraints` parses cleanly and panics inside `prove`. Not introduced here.

**Context:** `size` and `constraints` are read as independent header fields (prover.rs:180-186) and stored unvalidated. `prove_inner` allocates `vec![BlsScalar::zero(); self.size]` and writes it at every constraint index (prover.rs:324, :334), so `size = 0` on an otherwise valid blob indexes out of bounds; with public inputs, `Composer::dense_public_inputs` panics the same way one step earlier. `Composer::prove` does check `constraints` against the circuit, which is what makes the gap visible — nothing does the equivalent for `size`.

`prover_key.n` is loose in the same way: `quotient_poly` iterates a domain derived from `constraints` while indexing evaluation sets sized `8 * prover_key.n` — `v_h_coset_8n()[i]`, and `q_m.1[index]` and its siblings inside each widget's `compute_quotient_i`.

**Severity-rationale:** Pre-existing, and the contract the new tests assert does hold — `try_from_bytes` itself no longer panics. Raised because it is the remaining half of the same panic class, and the parser already holds both values.

**Fix:** In `try_from_bytes`, reject a header where `size != constraints.next_power_of_two()` or where `prover_key.n != size`; extend the new inner-key test with a tampered `size`.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `Composer::component_truncate<N>` to extract the low `N` bits of a witness [#867]
 - Add `Composer::component_range_bits<BITS>` range check counting bits directly [#867]
+- Add the public `Error::JubJubGeneratorNotPrimeOrder` variant [#867]
```
moCello: **NOTE:** Both new entries reference `[#867]`, the closed logic-gadget issue; this work traces to `[#832]`, which has no link definition in the file.

**Context:** The same reference is on the `component_mul_generator` entry at CHANGELOG.md:26-29. `[#867]` is right for the neighbouring `component_truncate`, `component_range_bits` and logic-gadget lines, so only these two need changing.

**Fix:** Change both to `[#832]` and add the link definition alongside the others at the bottom of the file.

moCello (approved):
APPROVE — the prover-key deserialization path is bounded end to end now: checked slicing, `try_from` on both length fields, a canonical-domain reconstruction in `Evaluations::from_slice`, and an `8n` domain check per evaluation set, all reached by tests through the public `try_from_bytes` entry. Valid keys still round-trip, and every debugger gate identity matches its widget source term for term. Nothing gates the merge.

Two notes post inline, both small. Three further points, none worth a round on its own, live only here:

- `evaluates_all_gate_identities` cannot fail for a mistranscribed identity. Its six positive rows are all zeros, where every identity vanishes however it is written, and its negative cases reach 5 of the ~16 identity slots — never `delta_xor_and`, `check_bit_consistency`, or the four accumulator identities, and `logic_xor` has no negative case at all. The transcriptions are correct at this head; the test is what would not say so. Worth a nonzero satisfying row per gate family and one negative case per slot next time the debugger is touched.
- `src/proof_system/widget.rs:327` still reads `v_h_coset_4n` for the `v_h_coset_8n` field — a stale marker of the kind this PR is clearing, in the file it hardened.
- The two notes from the previous round — the last-row rotated-wire read and the identity-only generator rejection test — are unchanged and still yours to take or leave; both are already posted inline, so they are not repeated here.

PR #875, opened 2026-07-29
Guard fixed-base leading-zero rounds at compile time

### LINKED ISSUE #874: Add compile-time guard for fixed-base leading-zero rounds
#871 closed the modulus-wrap soundness hole in `component_mul_generator` by forcing the top signed digits to zero, turning the closing scalar check into an integer equality. That argument rests on the effective signed-digit width — `FIXED_BASE_SIGNED_DIGIT_ROUNDS - FIXED_BASE_LEADING_ZERO_ROUNDS` — staying at or below 254 bits, and nothing checks it.

Two silent failure shapes if a future width edit moves the constants. At zero leading-zero rounds, the `leading_accumulator` capture coincides with the first-round witness already pinned to zero, so the closing `assert_equal_constant(leading_accumulator, zero)` degenerates to `0 == 0`. And already at width 255 the wrap becomes representable again: the accumulator endpoint can reach the claimed scalar plus the BLS modulus (`2^255 - 1` exceeds `q - (r - 1)`), and since `q < 2^255` a single forced-zero top digit is still satisfiable — so requiring merely one leading zero round is not enough. A negative constant is a const-eval compile error; these width configurations are the ones that compile clean.

One-line fix next to the constants: `const _: () = assert!(FIXED_BASE_SIGNED_DIGIT_ROUNDS - FIXED_BASE_LEADING_ZERO_ROUNDS <= 254);` with a short comment naming the bound it protects. The literal 254 is a property of the current field pair — the 252-bit Jubjub order against the 255-bit BLS scalar modulus — so a runtime test should derive the bound from the actual moduli and pin it: `2^254 - 1 < q - (r - 1)` (no wrap at the allowed width), `2^255 - 1 >= q - (r - 1)` (one more bit would reach one), and effective width at least 253 for honest-NAF completeness.

*(Edited: the original text proposed asserting `FIXED_BASE_LEADING_ZERO_ROUNDS >= 1`, which still admits the unsafe width-255 configuration; the width-form bound above is the invariant the soundness argument actually needs.)*

moCello:
Resolves #874

### DIFF
--- a/src/composer.rs
+++ b/src/composer.rs
@@ -103,6 +103,37 @@ const FIXED_BASE_SIGNED_DIGIT_ROUNDS: usize = 256;
 const FIXED_BASE_LEADING_ZERO_ROUNDS: usize =
     FIXED_BASE_SIGNED_DIGIT_ROUNDS - (JUBJUB_SCALAR_BITS + 1);
 
+/// Widest effective signed-digit width at which the fixed-base accumulator
+/// endpoint cannot encode a modulus wrap: `2^254 - 1 < q - (r - 1)`, verified
+/// against the actual moduli by the width-bound test in
+/// `fixed_base_soundness_tests`. Specific to the `(q, r)` pair — unrelated to
+/// the 254-bit caps of the logic and truncation gadgets, which derive from
+/// `q < 2^255` alone.
+const FIXED_BASE_MAX_SOUND_WIDTH: usize = 254;
+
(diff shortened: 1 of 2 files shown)


xevisalle (approved):
LGTM

HDauven (approved):
LGTM

The compile-time guards now cover both sides of the invariant and the regression test properly pins the width assumptions to the actual moduli

PR #876, opened 2026-07-30
Reject zero-Z fixed-base generators without panicking

HDauven:
`component_mul_generator` validates caller-supplied extended points with `is_on_curve`, but that method converts to affine and panics when `z = 0`.

This rejects zero-z representations before invoking the remaining point checks and adds a regression test confirming malformed generators return `JubJubGeneratorNotPrimeOrder`.

This is a host-side validation change only and does not alter the circuit layout.

### DIFF
--- a/src/composer.rs
+++ b/src/composer.rs
@@ -612,7 +612,11 @@ impl Composer {
     ) -> Result<WitnessPoint, Error> {
         let generator = generator.into();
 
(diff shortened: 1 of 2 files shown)


xevisalle (approved):
Maybe just add a short comment above the guard explaining why the apparently redundant Z check must precede `is_on_curve()`. Other than that, lgtm!

PR #878, opened 2026-07-31
Return a distinct error when proving an unsatisfied circuit

### LINKED ISSUE #877: Circuit-unsatisfiability failures surface only as PolynomialDegreeTooLarge
When proof generation fails because the circuit's constraints aren't satisfied, the prover always returns `Error::PolynomialDegreeTooLarge` — regardless of which gate is unsatisfied, or that the cause is unsatisfiability rather than a genuine degree/sizing problem. The failure is structural: an unsatisfied constraint makes the quotient polynomial fail to divide evenly, so it only surfaces during quotient commitment, after the which-constraint information is gone.

A circuit author debugging a failing proof gets no signal pointing at the offending gate. It would help to distinguish the ways a circuit can fail to prove — at minimum separating unsatisfiability from real degree/sizing errors, and ideally naming the unsatisfied constraint (the `debug`-feature debugger already evaluates each gate identity). Approach open; the prover's happy path is hot, so any always-on cost needs weighing.

moCello:
## Summary

Proving an unsatisfied circuit now returns a dedicated `Error::CircuitUnsatisfied` instead of the opaque `PolynomialDegreeTooLarge`, via an O(1) quotient-degree check; genuine sizing errors keep `PolynomialDegreeTooLarge`. Under the `debug` feature the debugger also names the first unsatisfied constraint on stderr.

Resolves #877

### DIFF
--- a/src/proof_system/quotient_poly.rs
+++ b/src/proof_system/quotient_poly.rs
@@ -99,8 +99,39 @@ pub(crate) fn compute(
         .collect();
 
     let coset = domain_8n.coset_ifft(&quotient);
+    let quotient_poly = Polynomial::from_coefficients_vec(coset);
 
-    Ok(Polynomial::from_coefficients_vec(coset))
+    // A satisfied assignment yields a numerator divisible by the vanishing
+    // polynomial of the domain, and the quotient's degree is bounded by the
+    // numerator's: the permutation product z(x) (degree n + 2, with hiding
+    // degree 2) times the four wire factors (degree n + 1 each, with hiding
+    // degree 1) has degree 5n + 6, and every gate identity stays below that.
+    // Dividing by the vanishing polynomial (degree n) leaves at most 4n + 6.
+    //
+    // An unsatisfied assignment leaves a nonzero remainder r(x) with
+    // deg r < n. On the 8n-sized coset the pointwise division then computes
+    // q(x) + r(x)/z_H(x), and 1/z_H interpolates over that coset to
+    // c_0 + c_1·x^n + ... + c_7·x^7n with c_7 nonzero (z_H only takes 8
+    // distinct values there), pushing the interpolated result to a degree of
+    // at least 7n. The check is anchored on that detection floor rather than
+    // the honest ceiling, so it stays correct should blinding or a new gate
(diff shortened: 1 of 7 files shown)


src/proof_system/quotient_poly.rs
```diff
@@ -99,8 +99,32 @@ pub(crate) fn compute(
...
+    // c_0 + c_1·x^n + ... + c_7·x^7n with c_7 nonzero (z_H only takes 8
+    // distinct values there), pushing the interpolated result to a degree of
+    // at least 7n. The check is anchored on that detection floor rather than
+    // the honest ceiling, so it stays correct should blinding or a new gate
+    // ever push the honest quotient past 4n + 6 — up to 7n of headroom.
+    // Unsatisfied assignments are caught even when the commit key is roomy
+    // enough (small circuits, deserialized oversized keys) that the split
+    // quotient chunks would have committed fine and only failed at
+    // verification.
+    if quotient_poly.degree() >= 7 * domain.size() {
```
HDauven: `.degree` first calls `is_zero()`, which scans the complete vector and then searchers backwards for the last non zero coefficient. `from_coefficients_vec` should already remove every trailing zero, so the extra full traversal is unnecessary I think.

Equivalent constant-time check is: 
```rust
if quotient_poly.len() > 7 * domain.size() {
    return Err(Error::CircuitUnsatisfied);
}
```

HDauven (changes requested):
There's one issue in the proving hot path. The new check is described as O(1), but the `Polynomial::degree` call scans the full coefficient vector if I'm not mistaken.

src/proof_system/quotient_poly.rs
moCello: Good catch — `degree()` is O(n) (`is_zero()` scans, then the backward loop). `quotient_poly` comes from `from_coefficients_vec`, which trims trailing zeros, so `len() == degree + 1` and `degree() >= 7n` is exactly `len() > 7n` (the empty/all-zero case is false under both). Switched to your form and updated the comment to justify the constant-time read. Suite unchanged and green.

HDauven: I think the savings are marginal in most cases but still, simpler flow

HDauven (approved):
Looks sound to me

PR #881, opened 2026-07-31
Split composer.rs into per-gadget submodules

### LINKED ISSUE #880: Reduce the size of composer.rs
`src/composer.rs` has grown to around 1935 lines, mixing the core composer machinery with several method clusters in one file, which makes it hard to navigate.

Extract its cohesive clusters into submodules — one or several — to bring the file size down. Pure refactor: no behaviour or public-API change intended. The boundaries, the number of submodules, and whether it lands as one PR or several are left to implementation.

moCello:
## Summary

Extract the gadget families from the ~1935-line `src/composer.rs` into seven submodules under `src/composer/` (`point`, `fixed_base`, `logic`, `range`, `truncate`, `bits`, `select`), leaving the composer machinery proper (~492 lines). Pure code motion — one commit per family; public API and behaviour are identical (methods stay on `impl Composer`), the gate-digest goldens are unchanged, and the only non-motion is four `pub(super)` visibility widenings forced by the cut.

Resolves #880


HDauven (approved):
LGTM

PR #882, opened 2026-08-03
Test the curve-addition helper-wire binding

moCello:
## Summary

Adds a soundness regression proving the curve-addition widget's `x_1 * y_2` helper wire (the `xy_consistency` term) cannot be forged: a witness that satisfies every other residual but assigns a wrong value to that wire is rejected. Both addends are public, so the gadget's intermediates are the prover's only freedom; the widget's output checks are solved for what a forged wire forces, leaving the binding term as the only objection, and a widget-level pin asserts that against the widget's own identity rather than a copy.

Also corrects `Polynomial::from_coefficients_vec`'s doc: it documented a panic on empty input that the implementation does not perform (empty or all-zero coefficients give the zero polynomial). Separate commit.

### DIFF
--- a/src/composer/point/soundness_tests.rs
+++ b/src/composer/point/soundness_tests.rs
@@ -4,15 +4,44 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-//! Soundness regressions for `assert_torsion_free_point`.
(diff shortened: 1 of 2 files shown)


HDauven (approved):
LGTM. The regression test looks solid, same with the polynomial documentation fix.

PR #883, opened 2026-08-03
Wrap the debugger's shifted wire reads to row 0

moCello:
## Summary

The `debug`-feature debugger read a rotated (`_w`) wire past the last constraint as zero, but the prover's evaluation domain is cyclic and wraps the last row's shifted wires to row 0. On an exact-power-of-two gate count (no padding) these differ. This wraps the shifted index modulo the padded circuit size so the debugger mirrors the prover regardless of how row 0 is populated.

No circuit built through the public API changes output — every proving path routes through `Composer::initialized()`, whose row 0 wires are all the zero witness, so wrapping to row 0 yields what the zero read already did. The change makes a diagnostic that is accidentally correct correct by construction; a new invariant test pins the row-0 property the correctness depends on. Four tests cover the exact-fill wrap, its silent-miss mirror, the padded-reads-zero case, and the invariant; `debug`-feature only, no production proving path touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   then equalled public zero modulo `q`, while the same digits drove the point
   endpoint to the nonidentity point `[q]G`, allowing a proof of a false public
   scalar/point relation.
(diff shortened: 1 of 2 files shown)


HDauven (approved):
LGTM

PR #885, opened 2026-08-05
Regroup the composer soundness tests into a dedicated subtree

moCello:
## Summary

The composer's soundness tests followed two layouts: five nested under their gadget as `composer/{family}/soundness_tests.rs`, and the sixth plus the shared support module flat at `composer/` level. Finding a gadget's coverage, or reading what the suite covers as a whole, meant knowing which layout applied.

This gathers all seven under `composer/tests/soundness/`, one module per family. It stays an in-crate module tree, not the crate-root `tests/` directory: these are unit tests exercising private internals, which the integration directory cannot reach at all.

Pure motion — every test body is unchanged, including the gate-digest goldens. Only the import blocks differ, since `use super::*` on a gadget module has to become the named imports it was standing in for.

The tests reach the gadgets' constraint emitters and width constants directly: a forgery only proves something when it emits the same gate layout as the honest gadget, which means calling the same emitter and sharing the same bounds. Nesting each family's tests under its gadget is what gave them that access, so relocating them widens exactly the items they use to `pub(super)` — reaching the `composer` subtree and nothing beyond it. `select_identity_gates` stays private; no test calls it.

### DIFF
--- a/src/composer.rs
+++ b/src/composer.rs
@@ -28,10 +28,7 @@ mod select;
 mod truncate;
 
 #[cfg(test)]
-mod evaluated_output_soundness_tests;
-
-#[cfg(test)]
-mod soundness_support;
+mod tests;
 
 pub(crate) mod permutation;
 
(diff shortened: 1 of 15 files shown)


xevisalle (approved):
LGTM

HDauven (approved):

PR #886, opened 2026-08-05
Make Error non_exhaustive and drop its Copy derive

### LINKED ISSUE #884: Make Error non_exhaustive
Error is a plain public enum, so every new variant is a breaking change for any downstream crate matching on it exhaustively — three variants were added this cycle alone, and more are queued. Marking it `#[non_exhaustive]` forces downstream matches to carry a wildcard arm, which stops future variants from breaking anyone. The attribute is itself a breaking change, so it should land while the current cycle's breaking window is open.

**Public API change:** `Error` becomes `#[non_exhaustive]` and loses its `Copy` derive. Downstream crates matching on `Error` exhaustively must add a wildcard arm; code relying on `Error` being `Copy` must clone instead. Both are breaking and land in the current major release window.

moCello:
## Summary

`Error` gains `#[non_exhaustive]`, so adding a variant stops being a breaking change for downstream crates. Three variants were added this cycle alone and more are queued.

The `Copy` derive comes off in the same change. It constrains every future variant to a payload that is itself `Copy`, and removing it later would need its own major release window.

Both are breaking: downstream matches on `Error` now need a wildcard arm, and code copying an `Error` out of a shared reference has to clone it. They land now because the release already carries breaking changes.

Closes #884

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Mark `Error` as `#[non_exhaustive]` [#884]
 - Change `component_add_point`, `component_sub_point` and `component_mul_point` to take and return `TorsionFreeWitnessPoint` [#870]
 - Change `component_neg_point` and `component_select_identity` to take and return `TorsionFreeWitnessPoint` [#870]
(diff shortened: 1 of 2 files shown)


xevisalle (approved):
lgtm

HDauven (approved):

PR #888, opened 2026-08-05
Validate KZG opening keys before use

### LINKED ISSUE #887: Harden KZG opening-key validation
## Summary

KZG opening keys reconstructed from bytes or archived public parameters should be validated before use. Their affine points are security-critical inputs, while the RKYV representation also stores derived prepared pairing values that can diverge from their affine sources.

## Expected behavior

- Checked deserialization rejects degenerate opening-key points.
- Archived public parameters revalidate the source points and rebuild prepared pairing values before producing verifier keys.
- Parameter generation cannot produce degenerate opening-key components.
- Existing serialized parameters and downstream Rusk/Phoenix artifacts remain compatible.

## Validation

- Add regressions for the compressed and RKYV parameter paths.
- Exercise downstream Phoenix artifact loading, fresh proof generation, verification, and transaction execution.

HDauven:
Resolves #887


src/commitment_scheme/kzg10/key.rs
```diff
@@ -321,21 +321,36 @@ impl Serializable<{ G1Affine::SIZE + G2Affine::SIZE * 2 }> for OpeningKey {
         let h = G2Affine::from_reader(&mut buffer)?;
         let beta_h = G2Affine::from_reader(&mut buffer)?;
 
-        Ok(Self::new(g, h, beta_h))
+        Self::try_new(g, h, beta_h)
     }
 }
 
 impl OpeningKey {
-    pub(crate) fn new(g: G1Affine, h: G2Affine, x_h: G2Affine) -> OpeningKey {
```
xevisalle, line 329: this only rejects identity points. However, RKYV’s derived CheckBytes validation is structural; it does not guarantee that archived affine points are on-curve and in the prime-order subgroup.

Consequently, src/commitment_scheme/kzg10/key.rs:352, called from src/commitment_scheme/kzg10/srs.rs:186, can accept malformed RKYV points and construct pairing state from them. This does not fully satisfy issue #887, which explicitly requires archived source points to be revalidated.

Suggested fix: make try_new require all three properties:

```rust
  let g_valid = bool::from(g.is_on_curve())
      && bool::from(g.is_torsion_free())
      && !bool::from(g.is_identity());

  let h_valid = bool::from(h.is_on_curve())
      && bool::from(h.is_torsion_free())
      && !bool::from(h.is_identity());

  let x_h_valid = bool::from(x_h.is_on_curve())
      && bool::from(x_h.is_torsion_free())
      && !bool::from(x_h.is_identity());

  if !(g_valid && h_valid && x_h_valid) {
      return Err(dusk_bytes::Error::InvalidData);
  }
```

Add RKYV regressions using off-curve and non-prime-order points, not only identity points.

xevisalle (changes requested):
only one comment about point validation, which has high impact I'd say. The rest looks sound to me.

xevisalle (approved):
lgtm!

src/commitment_scheme/kzg10/key.rs
```diff
@@ -321,21 +321,43 @@ impl Serializable<{ G1Affine::SIZE + G2Affine::SIZE * 2 }> for OpeningKey {
...
+    ) -> Result<OpeningKey, dusk_bytes::Error> {
+        let g_valid = bool::from(g.is_on_curve())
+            && bool::from(g.is_torsion_free())
+            && !bool::from(g.is_identity());
+        let h_valid = bool::from(h.is_on_curve())
+            && bool::from(h.is_torsion_free())
+            && !bool::from(h.is_identity());
+        let x_h_valid = bool::from(x_h.is_on_curve())
+            && bool::from(x_h.is_torsion_free())
+            && !bool::from(x_h.is_identity());
```
moCello, line 342: **NOTE:** On the rkyv path these predicates run against archived field limbs nothing has checked for canonicity, and `try_new` stores those same limbs into the key the verifier pairs with.

**Context:** `dusk_bls12_381::Fp` is a bare `[u64; 6]` behind a derived `CheckBytes` (`dusk-bls12_381-0.14.2/src/fp.rs:21-27`), so an archived point can carry any 384-bit limb pattern — `off_curve_g1()` in this PR relies on exactly that. `Fp::mul` reduces via `montgomery_reduce`, whose output bound holds only for inputs below the modulus. Limbs equal to a canonical value plus `p` stay value-correct and are harmless; whether limbs far enough above `p` can pass both predicates while pairing as a different point is not settled either way.

**Severity-rationale:** Nothing reaches `PublicParameters` through rkyv today, and that path carried no validation at all before this change — an under-closed pre-existing gap, not one the change opens.

**Fix:** Canonicalize through the compressed encoding in `validated()` — `G1Affine::from_bytes(&g.to_bytes())` and the G2 equivalent — and store the round-tripped points. `Fp::to_bytes` reduces any limb pattern to the canonical encoding of the value it represents, so the round-trip is lossless on honest input, and it routes canonicity, on-curve and subgroup membership through one audited decoder. Keeping it out of `try_new` avoids a second decode on the byte path, where `from_bytes` already performs all three checks.

src/commitment_scheme/kzg10/srs.rs
```diff
@@ -191,7 +190,11 @@ impl PublicParameters {
         let truncated_prover_key = self
             .commit_key
             .truncate(truncated_degree + Self::ADDED_BLINDING_DEGREE)?;
-        let opening_key = self.opening_key.clone();
+        // `PublicParameters` can be reconstructed through rkyv, which archives
+        // affine and prepared G2 values independently. Revalidate the affine
+        // source values and rebuild the prepared caches before they become
+        // usable by a verifier.
+        let opening_key = self.opening_key.validated()?;
```
moCello, line 196: **NOTE:** Archived opening keys stay unvalidated until `trim` cleans them, so the soundness of the rkyv path rests on every consumer routing through `trim` — a comment, not the type system.

**Context:** `rkyv::from_bytes::<PublicParameters>` still succeeds on a key whose points are off-curve, out-of-subgroup or the identity, and whose `prepared_h`/`prepared_x_h` are attacker-chosen; the new tests encode that, `.expect(...)`-ing the deserialization at `srs.rs:326` and `srs.rs:352` and asserting only on `trim`.

**Severity-rationale:** `opening_key` and `trim` are both `pub(crate)`, and `Compiler::compile` routes through `trim`, so no consumer can hold the unvalidated key today. An accessor added later re-opens the vacuous-pairing hole with no compile-time signal.

**Fix:** Reduce the archived representation to the three affine points and rebuild through `try_new` when converting out of the archive — then no archived state can bypass the predicates, and the prepared caches stop being archived at all. Hand-writing `impl CheckBytes for ArchivedOpeningKey` with the point predicates is the alternative if the archive format must keep its shape (`ArchivedVerifierKey`, `src/proof_system/widget.rs:63`, and `ArchivedProof`, `src/proof_system/proof.rs:103`, show the hand-written form, though theirs only delegate structural checks via `util::check_field`). `trim`'s rebuild stays either way as the backstop.

src/commitment_scheme/kzg10/key.rs
```diff
@@ -321,21 +321,43 @@ impl Serializable<{ G1Affine::SIZE + G2Affine::SIZE * 2 }> for OpeningKey {
...
             x_h,
             prepared_h,
             prepared_x_h,
-        }
+        })
+    }
+
+    pub(crate) fn validated(&self) -> Result<OpeningKey, dusk_bytes::Error> {
+        Self::try_new(self.g, self.h, self.x_h)
     }
```
moCello, line 365: **NOTE:** The two archived `G2Prepared` fields are now dead weight — ~39 KB per archived `PublicParameters` — and removing them breaks the archive format, so this is a follow-up rather than this PR.

**Context:** Nothing reads the archived copies once the prepared values are always recomputed here. Each field archives exactly 68 `Fp2` triples (`dusk-bls12_381-0.14.2/src/pairings.rs:556-561`), 19,584 bytes apiece, and carries an unchecked length: rkyv validates the archived vector's layout, not its element count, while `multi_miller_loop` indexes `coeffs[index]` up to 67 before the identity select can discard the term (`pairings.rs:590`). That surface is unreachable while nothing reads the archived copies, and only while that holds.

**Fix:** Skip the two fields in the archived representation — the payload and the unchecked length go together.

src/commitment_scheme/kzg10/key.rs
```diff
@@ -321,21 +321,43 @@ impl Serializable<{ G1Affine::SIZE + G2Affine::SIZE * 2 }> for OpeningKey {
         let h = G2Affine::from_reader(&mut buffer)?;
         let beta_h = G2Affine::from_reader(&mut buffer)?;
 
-        Ok(Self::new(g, h, beta_h))
+        Self::try_new(g, h, beta_h)
```
moCello, line 324: **NOTE:** `from_bytes` becoming fallible for canonically-encoded identity points turns `PublicParameters::from_slice_unchecked` into a process abort and leaves its "no checks" doc stale.

**Context:** `srs.rs:133-134` calls `OpeningKey::from_slice(...).expect("Error at OpeningKey deserialization")`. Only malformed compressed encodings failed before; `G1Affine::identity().to_bytes()` is a well-formed encoding that decoded fine and now returns `InvalidData`. The doc at `srs.rs:121-123` still states no checks are performed on the points the bytes contain.

**Severity-rationale:** The `expect` predates this change and the function is `unsafe` with a trusted-source contract; what widened is the class of inputs that reach it.

**Fix:** State in the doc that the opening key is always validated while the commit key is not, or return the error instead of panicking.

src/commitment_scheme/kzg10/srs.rs
```diff
@@ -251,4 +315,59 @@ mod test {
...
+        pp.opening_key.prepared_x_h = G2Prepared::from(G2Affine::identity());
+
+        let bytes = rkyv::to_bytes::<_, 256>(&pp).expect("archive parameters");
+        let decoded = rkyv::from_bytes::<PublicParameters>(&bytes)
+            .expect("validate and deserialize parameters");
+        let (_, opening_key) = decoded.trim(2).expect("trim parameters");
+
+        let point = BlsScalar::from(7u64);
+        let proof = super::super::proof::Proof {
+            commitment_to_witness: super::super::Commitment::default(),
```
moCello: **NOTE:** `trim_rebuilds_archived_prepared_points` claims more than it checks: only `prepared_h` is exercised, and the `is_err()` shape cannot cover `prepared_x_h` at all.

**Context:** `Commitment::default()` is the identity, so `total_w` is the identity and `batch_check`'s first pairing term `(affine_total_w, prepared_x_h)` is discarded by `multi_miller_loop`'s `term.0.is_identity()` branch whatever `prepared_x_h` holds. The `is_err()` shape is what blocks isolation in general: leaving a prepared value archived only ever *drops* its pairing term, so on a proof that should fail, the surviving honest term satisfies the assertion on its own.

**Fix:** Narrow the message to name `prepared_h`, or — to cover both — assert `is_ok()` on an honest opening proof with one prepared field corrupted per case. Only that direction fails when a term is silently dropped.

src/commitment_scheme/kzg10/srs.rs
```diff
@@ -206,10 +209,71 @@ impl PublicParameters {
...
+    #[cfg(feature = "rkyv-impl")]
+    fn assert_archived_opening_key_rejected(
+        mutate: impl FnOnce(&mut OpeningKey),
+    ) {
+        let mut pp = PublicParameters::setup(8, &mut OsRng).unwrap();
+        mutate(&mut pp.opening_key);
+
+        let bytes = rkyv::to_bytes::<_, 256>(&pp).unwrap();
+        let decoded = rkyv::from_bytes::<PublicParameters>(&bytes).unwrap();
+        assert!(decoded.trim(2).is_err());
```
moCello: **NOTE:** The rkyv rejection assertions accept any error from `trim`, including one raised before the validation they are testing runs.

**Context:** `trim` calls `commit_key.truncate(...)` before `opening_key.validated()`, so a `TruncatedDegreeTooLarge` satisfies these assertions as well as the point rejection does. The current degrees leave headroom, so it cannot fire today — but nothing enforces that, and a later change to them would leave the tests green while testing nothing. `srs.rs:330` has the same shape.

**Fix:** Match on the expected error, as `tests/opening_key_validation.rs:133-136` does.

src/commitment_scheme/kzg10/srs.rs
```diff
@@ -251,4 +315,59 @@ mod test {
...
+        pp.opening_key.prepared_h = G2Prepared::from(G2Affine::identity());
+        pp.opening_key.prepared_x_h = G2Prepared::from(G2Affine::identity());
+
+        let bytes = rkyv::to_bytes::<_, 256>(&pp).expect("archive parameters");
+        let decoded = rkyv::from_bytes::<PublicParameters>(&bytes)
+            .expect("validate and deserialize parameters");
+        let (_, opening_key) = decoded.trim(2).expect("trim parameters");
+
+        let point = BlsScalar::from(7u64);
+        let proof = super::super::proof::Proof {
```
moCello: **NOTE:** Fully qualified `super::super::` paths inside a function body; the file keeps every `use` at module top and refers to items by short name.

**Fix:** Add `use super::super::Commitment;` and `use super::super::proof::Proof;` to the test module's import block, gated on `rkyv-impl` alongside the `G2Prepared` and `Transcript` imports already there.

moCello (approved):
APPROVE — the change closes #887 on both paths it names and nothing gates the merge. All three points are checked for on-curve, subgroup membership and non-identity, the prepared pairing values are rebuilt from the validated sources rather than trusted from the archive, and both entry points are covered by tests. No public API moves and no compatibility break — the byte path's decoder already ran the on-curve and subgroup checks, so only the identity rejection is new there — and the added cost lands in `trim`, which runs once per circuit compile.

Seven notes below, none blocking and most of them small. The two worth an answer are the rkyv ones: neither is reachable by any consumer today, and both leave that path better than it was before this change, so they are follow-up shaped rather than merge-shaped.

moCello (approved):
APPROVE — All prior-round findings are resolved: every path that reconstructs an opening key routes through the same three-predicate validation, the archived-parameter path revalidates canonical encodings and rebuilds the prepared pairing values in `trim`, and the regression suite rejects each degenerate point class for the specific reason under test while honest artifacts keep verifying. The two remaining architecture points — validation living in `trim` rather than the type system, and the archived prepared fields being dead weight — would break the archive format if addressed here; they are follow-up material and do not gate this change.

PR #891, opened 2026-08-06
Validate point inputs before projecting them to affine

### LINKED ISSUE #889: Validate point inputs before conversion in the point gadgets
`append_point`, `append_constant_point`, `append_public_point` and `assert_equal_public_point` all accept anything convertible into an affine point, and converting a degenerate extended point (z = 0) panics inside the conversion rather than returning an error. `component_mul_generator` was guarded against the same input in an earlier change; these four were not.

Narrowing the signatures would only move the panic into callers, since no checked conversion exists to offer them, so the fix is to validate before converting and return an error.

This is a breaking change to four public signatures and is intended for the current major release window.

**Second break:** the bound also moves from `impl Into<JubJubAffine>` to `impl Into<JubJubExtended>`, which no longer accepts a reference — the old bound took `&JubJubExtended` through jubjub's `From<&ExtendedPoint> for AffinePoint`, and there is no equivalent for the extended target. Call sites passing a reference, as the key and signature accessors across the ecosystem return one, need a `*` alongside the new `?`. `JubJubExtended` is `Copy`, so each fix is one character and the compiler points at every site.

moCello:
## Summary

`append_point`, `append_constant_point`, `append_public_point` and `assert_equal_public_point` accepted anything convertible into an affine point and did the conversion first. A `JubJubExtended` with `Z = 0` denotes no affine point, so the projection inverts zero and aborts the process — and nothing upstream of the four established that `Z` was nonzero.

All four now take `impl Into<JubJubExtended>` and guard `Z` before projecting, the shape `component_mul_generator` already uses. Narrowing to `JubJubAffine` would only have moved the same unwrap into callers, since no checked conversion exists to offer them.

`append_constant_point`'s membership check now runs on the extended point rather than on the projection, which additionally ties `T1 · T2` to the affine coordinates, so an inconsistent extended representation of an on-curve subgroup point is refused. The other three establish no membership, unchanged.

Two breaks for consumers: the new `Result`, and the bound no longer accepting a reference — call sites passing one need a `*`. Both are recorded in the CHANGELOG.

A second commit trims the unreleased changelog entries. They had grown into prose, restating consequences and rationale that the linked issues already carry, and each is reduced to what changed plus its issue link. That reaches beyond this change: entries belonging to #867, #870, #877 and #884 are shortened too, and two entries carrying two facts each are split rather than compressed. Entries contributed by others are left untouched.

Closes #889

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,38 +9,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add `Composer::assert_torsion_free_point` boundary check constraining a witness point into the prime-order subgroup, returning it as `TorsionFreeWitnessPoint` [#870]
-- Add `TorsionFreeWitnessPoint` encoding prime-order subgroup membership in the type system [#870]
-- Add `Error::JubJubPointNotTorsionFree` for constant points outside the prime-order subgroup [#870]
-- Add `Error::JubJubPointDegenerate`, returned by the point entry points for an extended point with a zero `Z` coordinate [#889]
-- Add `Composer::component_truncate<N>` to extract the low `N` bits of a witness [#867]
+- Add `Composer::assert_torsion_free_point` [#870]
+- Add `TorsionFreeWitnessPoint` [#870]
+- Add `Error::JubJubPointNotTorsionFree` [#870]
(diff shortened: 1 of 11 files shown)


HDauven (approved):
LGTM

PR #893, opened 2026-08-06
Rebuild KZG opening keys at the RKYV boundary

### LINKED ISSUE #890: Rebuild KZG opening keys at the RKYV boundary
## Summary

KZG opening keys currently archive both their affine source points and the derived `G2Prepared` caches. RKYV validates only the archived layout, so malformed source points can deserialize and remain unvalidated until `trim`. The prepared caches are also no longer trusted after #887 and add roughly 39 KB to every archived parameter set.

## Expected behavior

- Archive only the three affine source points.
- Canonicalize and validate those points while converting out of the archive.
- Rebuild `prepared_h` and `prepared_x_h` from the validated sources.
- Reject an invalid opening key before a consumer can obtain `PublicParameters`.

This changes the archive representation and should be handled in the next compatible release window.

HDauven:
Resolves #890

### DIFF
--- a/.github/workflows/dusk_ci.yml
+++ b/.github/workflows/dusk_ci.yml
@@ -52,6 +52,7 @@ jobs:
       - uses: dsherret/rust-toolchain-file@v1
       - uses: Swatinem/rust-cache@v2
       - run: make no-std
+      - run: cargo check --target thumbv6m-none-eabi --no-default-features --features alloc,rkyv-impl,rkyv/size_32
 
   test:
     name: Tests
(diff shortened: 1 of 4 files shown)


.github/workflows/dusk_ci.yml
```diff
@@ -52,6 +52,7 @@ jobs:
       - uses: dsherret/rust-toolchain-file@v1
       - uses: Swatinem/rust-cache@v2
       - run: make no-std
+      - run: cargo check --no-default-features --features alloc,rkyv-impl,rkyv/size_32,bytecheck/std
```
xevisalle: non-blocking, but, this enables bytecheck/std and checks the host target, so it is not a strict embedded no_std check. Maybe replace it with:

run: cargo check --target thumbv6m-none-eabi --no-default-features --features alloc,rkyv-impl,rkyv/size_32

xevisalle (approved):
just a nit

PR #894, opened 2026-08-07
Test that the fixed-base widget binds its xy_alpha wire

### LINKED ISSUE #892: Add a binding regression for the fixed-base widget's xy_alpha wire
The fixed-base scalar-mul widget carries an intermediate wire, `xy_alpha`, whose only in-identity binding is the `xy_consistency` residual — the two accumulator residuals consume it but each remain satisfiable for any value it takes.

That term is present and correct in the quotient computation, the prover's linearization, and the verifier key, but no test asserts it: a witness forging that wire has never been shown to be rejected, so a refactor could weaken the term with the suite staying green.

This adds a soundness regression that forges the wire, solves the accumulator checks for the output it forces, and requires the proof to fail — extending the coverage #882 added for the curve-addition widget's helper wire.

moCello:
## Summary

The fixed-base scalar-multiplication widget keeps its round formula below degree 4 by allocating a per-row witness for `x_alpha * y_alpha`. Its identity folds four residuals, and only one of them — `xy_consistency` — binds that wire; the two accumulator checks consume it in place of the product rather than constraining it. Both are linear in their own output coordinate, so at any wire value they solve for a matching accumulator, and nothing constrains that accumulator to the curve. A prover who forges the wire and carries the forced accumulator forward satisfies every other residual and claims a public multiple of their choosing. No test exercised that term.

This adds `fixed_base_binds_the_helper_wire`, which forges the wire four ways — off-by-one mid-chain, off-by-one and zero on the last round, and one solved so the claimed multiple's x-coordinate lands on `[7]G` — and requires each to fail with an unsatisfied constraint on the honest gate layout. Each runs against two scalars, one with a dense width-2 NAF and one with only two nonzero digits, so the mid-chain forgery is exercised both on a round that carries a digit and on one that does not.

Rejection on its own would say little, since bit consistency or the public-input equality could produce one while the term under test never objects. `forced_accumulators_satisfy_the_widget_output_checks` pins that down: it requires the widget's identity to reduce to the binding term at every round of every forged chain, under two challenges so the two accumulator residuals cannot cancel, and perturbs each forced coordinate to show its own check still constrains it. It does so at all three places the widget computes the identity — the prover's quotient, the prover's linearization, and the verifier key's — so a weakening cannot hide in one of them. `component_mul_generator_layout_matches_golden` pins the production layout separately, so the fork and the gadget cannot drift together.

The only production change is visibility and docs: `Composer::assert_canonical_jubjub_scalar` goes from private to `pub(super)` so the forgery calls the canonicality guard rather than carrying a second copy of its gates, matching `append_fixed_base_signed_digits`, which is already reachable from this suite for the same reason. `pub(super)` on an item in `composer::fixed_base` resolves to `pub(in crate::composer)`, so nothing leaves the crate, and the helper only ever adds constraints. No public API change and no gate moves, which the unchanged layout golden confirms.

Closes #892.

### DIFF
--- a/src/composer/fixed_base.rs
+++ b/src/composer/fixed_base.rs
@@ -153,10 +153,10 @@ impl Composer {
     /// Constrain a fixed-base multiplication for a supplied signed-digit
     /// witness assignment.
     ///
-    /// This is separated from honest wNAF generation so the soundness tests
-    /// can inject adversarial assignments into the exact production layout.
-    /// The constraints, not the provenance of `signed_digits`, must reject
-    /// noncanonical and field-wrapping representations.
+    /// Split from honest wNAF generation, and reachable from the `composer`
+    /// subtree for its unit tests. The constraints, not the provenance of
+    /// `signed_digits`, must reject noncanonical and field-wrapping
+    /// representations.
     pub(super) fn append_fixed_base_signed_digits(
         &mut self,
         jubjub: Witness,
@@ -312,7 +312,9 @@ impl Composer {
     /// `(r - 1) - scalar < 2^252`; if `scalar >= r`, that subtraction
     /// underflows modulo the much larger BLS scalar-field modulus and cannot
     /// satisfy the range check.
-    fn assert_canonical_jubjub_scalar(&mut self, scalar: Witness) {
+    ///
+    /// Reachable from the `composer` subtree for its unit tests.
+    pub(super) fn assert_canonical_jubjub_scalar(&mut self, scalar: Witness) {
         self.range_check(scalar, JUBJUB_SCALAR_BITS);
 
         let max_jubjub_scalar = BlsScalar::from(-JubJubScalar::one());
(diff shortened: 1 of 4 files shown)


HDauven (approved):
LGTM!

One non-blocking improvement we could still make is to harden the verifier-key pinning by using distinct synthetic commitments and asserting both the scalar and point vectors returned by `compute_linearization_commitment`. That would make the regression more self-contained against selector/commitment mixups in a future refactor, unless you deem it overkill.

moCello:
Good catch that the pin can't currently distinguish the three commitments. I'm inclined to leave it here: a commitment mixup is an inconsistent change, so it breaks the pairing check and every proof test with it — loud in a way the binding term isn't, which is the case this regression exists for. I'd rather spend it on the next widget's pin, where the same construction is being written from scratch and asserting both vectors costs nothing — tracked in #901.

PR #896, opened 2026-08-07
Implement Display and Error against core

### LINKED ISSUE #895: Provide Display and Error for the crate's Error type without std
The `Error` type and its prelude re-export are unconditional, but the `Display` and `std::error::Error` impls sit behind the `std` feature. A build without default features — which is what the bare-metal check exercises and what `dusk-core` consumes — therefore gets an error type it cannot print and cannot use as an `Error`.

Since `Error` became `#[non_exhaustive]`, downstream matches need a wildcard arm, and in `no_std` that arm has only `Debug` behind it.

Retargeting both impls at `core` is additive: `core::error::Error` is stable well below this crate's minimum Rust version.

moCello:
## Summary

`mod error` and the prelude re-export of `Error` are unconditional, but `Display` and the error-trait impl were both behind `#[cfg(feature = "std")]`. A `--no-default-features` build — what `make no-std` exercises and what `dusk-core` consumes — got an `Error` it could neither print nor use as an error. #884 sharpened this by marking the type `#[non_exhaustive]`, so every downstream `match` needs a wildcard arm that, in a no_std build, had only `Debug` behind it.

Both impls are now unconditional and target `core`. The `Display` body writes only integers and `Debug` arguments, all of which `core::fmt` provides, and `core::error::Error` has been stable since Rust 1.81 against the crate's `rust-version = "1.85"`. Since that release `std::error::Error` is a re-export of the `core` trait, so `std` consumers see the same impl surface and the same rendered strings as before; the test asserts that coercion so a future divergence fails to compile.

The test module's `feature = "std"` gate outlived its reason and is now `feature = "alloc"`, which is what its body actually needs — it renders through `to_string`.

Purely additive: no variants added, renamed, or reworded, no feature definitions touched, and the orphan rule means no consumer can hold a conflicting impl.

Resolves #895

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `Composer::component_range_bits<BITS>` range check counting bits directly [#867]
 - Add the public `Error::JubJubGeneratorNotPrimeOrder` variant [#832]
 - Add `Error::CircuitUnsatisfied` [#877]
+- Add `Display` for `Error` to builds without the `std` feature [#895]
+- Add `core::error::Error` for `Error` to builds without the `std` feature [#895]
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


HDauven (approved):
LGTM

PR #899, opened 2026-08-07
Reject zero vanishing-coset evaluations when decoding Provers

### LINKED ISSUE #898: Reject zero vanishing-coset evaluations when decoding Provers
## Summary

`ProverKey::from_slice` reconstructs and validates the domain and length of `v_h_coset_8n`, but accepts zero field elements in the cached evaluations.

`quotient_poly::compute` treats every cached value as a nonzero vanishing-polynomial denominator and unwraps its inverse. A single zero evaluation therefore passes checked `Prover` deserialization and deterministically panics during proving. In release builds, the crate uses `panic = "abort"`.

## Expected behavior

- Checked `Prover` deserialization rejects zero vanishing-coset evaluations.
- Malformed prover artifacts return an error before quotient construction.
- Valid serialized provers continue to round-trip and prove normally.

## Validation

- Add a regression that zeroes one `v_h_coset_8n` entry in an honest serialized prover and requires `Prover::try_from_bytes` to reject it.

HDauven:
## Summary

Reject zero cached vanishing-coset evaluations during checked `Prover` deserialization so quotient construction cannot try to invert zero and panic.

## Validation

- `make fmt`
- `make clippy`
- `make no-std`
- `make test`

Resolves #898

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Reject zero vanishing-coset evaluations during checked `Prover`
+  deserialization [#898]
 - Reject degenerate KZG opening keys and rebuild archived prepared pairing
   values from their validated affine points before use [#887]
 - Bind `append_logic_xor`/`append_logic_and` output to their inputs [#867]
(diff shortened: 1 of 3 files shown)


moCello (approved):
APPROVE — the check closes the panic path from #898 at the artifact boundary, regression tests cover both the inner decoder and the public deserialization entry, and honest prover keys cannot contain a zero vanishing evaluation, so valid artifacts keep round-tripping; nothing here needs changes before merge.

One follow-up worth its own PR rather than this one: the cached vanishing evaluations are fully derivable from the domain `from_slice` already reconstructs — each entry is `GENERATOR^n · ω^(n·i) − 1`, cycling through 8 distinct values — so validating the decoded table against the recomputed values (or replacing it outright) would reject every tampered entry at the same cost as the zero scan. As merged, a tampered nonzero entry decodes successfully and only surfaces as proofs that fail verification, with no error pointing at the corrupted artifact.

PR #900, opened 2026-08-07
Reject empty commitment keys when decoding Provers

### LINKED ISSUE #897: Reject empty commitment keys when decoding Provers
## Summary

`Prover::try_from_bytes` validates the encoded commitment-key length and every decoded point, but still accepts a structurally valid commitment key containing no points.

The resulting `Prover` reaches `CommitKey::max_degree` during proving. With overflow checks enabled, subtracting one from the empty key length panics. With overflow checks disabled, as in a default downstream release build, the subtraction wraps to `usize::MAX`, allowing the degree check to accept values the empty key cannot support and causing later proving operations to fail or panic.

Checked deserialization should reject the malformed artifact before it reaches either path.

## Expected behavior

- Checked `Prover` deserialization rejects an empty commitment key.
- Malformed prover artifacts return an error without panicking during decoding or later proving.
- Valid serialized provers continue to round-trip and prove normally.

## Validation

- Add a regression that replaces an honest prover's commitment key with a correctly encoded empty key and requires `Prover::try_from_bytes` to reject it.

HDauven:
## Summary

Reject empty commitment keys during checked `Prover` deserialization so malformed artifacts cannot reach `CommitKey::max_degree` and panic.

## Validation

- `make fmt`
- `make clippy`
- `make no-std`
- `make test`

Resolves #897

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Reject zero vanishing-coset evaluations during checked `Prover`
   deserialization [#898]
+- Reject empty commitment keys during checked `Prover` deserialization [#897]
 - Reject degenerate KZG opening keys and rebuild archived prepared pairing
   values from their validated affine points before use [#887]
 - Bind `append_logic_xor`/`append_logic_and` output to their inputs [#867]
@@ -739,6 +740,7 @@ is necessary since `rkyv/validation` was required as a bound.
 
 <!-- ISSUES -->
 [#898]: https://github.com/dusk-network/plonk/issues/898
+[#897]: https://github.com/dusk-network/plonk/issues/897
 [#895]: https://github.com/dusk-network/plonk/issues/895
 [#890]: https://github.com/dusk-network/plonk/issues/890
 [#889]: https://github.com/dusk-network/plonk/issues/889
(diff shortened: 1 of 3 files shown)


moCello (approved):
APPROVE — The guard sits in the layer that owns the invariant, `Prover::try_from_bytes` is its only caller, it cannot reject a key `PublicParameters::setup` can produce, and both regression tests fail without it. Nothing gates the merge; no inline comments follow.

Two points for after the merge, neither affecting this diff.

The same empty-key hazard remains on the rkyv boundary. `PublicParameters` derives `Archive`/`CheckBytes` with `commit_key: CommitKey` unwrapped, so `rkyv::from_bytes::<PublicParameters>` accepts an empty `powers_of_g`; `PublicParameters::max_degree` then computes `len() - 1` on it, and `Compiler::compile` → `trim` → `truncate` panics under either `overflow-checks` setting. That derive also admits off-curve commit-key points, which `from_raw_var_bytes` rejects. This is the field #887 left behind when it hardened `opening_key` via `OpeningKeyRkyv`. Nothing in the Dusk tree decodes `PublicParameters` through rkyv today, so the exposure is to external consumers of the published crate — out of scope for #897 and worth its own issue. The durable shape is `max_degree` returning `Option`/`Result` via `checked_sub(1)`, or one validating constructor that every `CommitKey` decode path routes through, rkyv included.

#897's stated failure mode overstates the exposure. `panic = "abort"` and `overflow-checks = false` live in this crate's `[profile.release]`, and Cargo reads profile tables only from the manifest of the package being built — a downstream release build sees neither. Under default release settings `len() - 1` on an empty key wraps instead of panicking, which makes `check_commit_degree_is_within_bounds` accept any degree rather than terminating the process. The fix and the changelog entry are right; the issue text is what merits the correction.

PR #904, opened 2026-08-07
Validate cached vanishing evaluations when decoding Provers

### LINKED ISSUE #902: Validate cached vanishing evaluations when decoding Provers
## Summary

`ProverKey::from_slice` validates the domain and rejects zero values in the cached vanishing-coset evaluations, but nonzero values are otherwise accepted as-is.

The complete table is derivable from the reconstructed domain and only cycles through eight distinct values. A corrupted nonzero table can therefore pass checked deserialization and only fail later during proving or verification, without identifying the prover artifact as the source of the error.

## Expected behavior

- Checked `ProverKey` deserialization verifies the cached vanishing-coset evaluations against the values derived from the domain, or reconstructs the table instead of trusting it.
- Any modified cached evaluation is rejected at the artifact boundary.
- Valid serialized provers continue to round-trip, prove, and verify normally.

## Validation

- Add a regression that changes one cached evaluation to another nonzero scalar and requires deserialization to reject it.
- Add a regression for a consistently modified nonzero table, such as scaling every entry.

HDauven:
## Summary

Validate cached vanishing-coset evaluations against the values derived from the reconstructed domain during checked `Prover` deserialization. The shared recurrence avoids allocating a second evaluation table and replaces the previous per-entry exponentiation used during compilation.

## Validation

- `make fmt`
- `make clippy`
- `make no-std`
- `make test`

Resolves #902

### DIFF
--- a/src/fft/domain.rs
+++ b/src/fft/domain.rs
@@ -295,21 +295,36 @@ pub(crate) mod alloc {
             &self,            // domain to evaluate over
             poly_degree: u64, // degree of the vanishing polynomial
         ) -> Evaluations {
-            assert!((self.size() as u64) > poly_degree);
-            let coset_gen = GENERATOR.pow(&[poly_degree, 0, 0, 0]);
-            let v_h: Vec<_> = (0..self.size())
-                .map(|i| {
-                    (coset_gen
-                        * self.group_gen.pow(&[
-                            poly_degree * i as u64,
-                            0,
-                            0,
-                            0,
-                        ]))
-                        - BlsScalar::one()
-                })
-                .collect();
+            let v_h = self.vanishing_poly_over_coset(poly_degree).collect();
             Evaluations::from_vec_and_domain(v_h, *self)
+        }
+
+        pub(crate) fn matches_vanishing_poly_over_coset(
+            &self,
+            poly_degree: u64,
+            evaluations: &[BlsScalar],
+        ) -> bool {
+            poly_degree < self.size() as u64
+                && evaluations.len() == self.size()
+                && evaluations
+                    .iter()
+                    .copied()
+                    .eq(self.vanishing_poly_over_coset(poly_degree))
+        }
+
+        fn vanishing_poly_over_coset(
+            &self,
+            poly_degree: u64,
+        ) -> impl Iterator<Item = BlsScalar> {
+            assert!((self.size() as u64) > poly_degree);
+            let mut point = GENERATOR.pow(&[poly_degree, 0, 0, 0]);
+            let step = self.group_gen.pow(&[poly_degree, 0, 0, 0]);
+
+            (0..self.size()).map(move |_| {
+                let evaluation = point - BlsScalar::one();
+                point *= step;
+                evaluation
+            })
         }
 
         /// Return an iterator over the elements of the domain.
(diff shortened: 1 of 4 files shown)


src/fft/domain.rs
```diff
@@ -295,23 +295,37 @@ pub(crate) mod alloc {
...
+                    .iter()
+                    .copied()
+                    .eq(self.vanishing_poly_over_coset(poly_degree))
+        }
+
+        fn vanishing_poly_over_coset(
+            &self,
+            poly_degree: u64,
+        ) -> impl Iterator<Item = BlsScalar> {
+            assert!((self.size() as u64) > poly_degree);
```
moCello, line 319: **NOTE:** Validation predicate for untrusted bytes panics via `assert!` instead of returning `false` if the domain/degree invariant breaks upstream — unreachable today, and a two-line change to close.

**Context:** The assert cannot fire from `ProverKey::from_slice` today — `n = 0` fails the power-of-two check, and for `n >= 1` the 8n domain always exceeds `poly_degree = n` — but that safety lives in the caller. `matches_vanishing_poly_over_coset` is precisely the function asked whether attacker-controlled data is well-formed; a later relaxation of the upstream checks would turn checked deserialization into a panic.

**Fix:** Make the predicate total: have `matches_vanishing_poly_over_coset` return `false` when `poly_degree >= self.size()` instead of reaching the assert, keeping the assert (or a `debug_assert!`) for the compile-path caller if desired.

src/proof_system/widget.rs
```diff
@@ -554,7 +554,12 @@ pub(crate) mod alloc {
             let perm_linear_evaluations = evals_from_reader(&mut buffer)?;
 
             let v_h_coset_8n = evals_from_reader(&mut buffer)?;
-            if v_h_coset_8n.evals.contains(&BlsScalar::zero()) {
+            let poly_degree =
+                u64::try_from(n).map_err(|_| dusk_bytes::Error::InvalidData)?;
+            if !evaluations_domain.matches_vanishing_poly_over_coset(
```
moCello, line 559: **NOTE:** `perm_linear_evaluations`, read a few lines above, is the same corruption class this check closes — domain-derivable yet accepted as-is; follow-up issue, not this PR.

**Context:** Compile builds `linear_evaluations` as `domain_8n.coset_fft(&[BlsScalar::zero(), BlsScalar::one()])` — the coset points themselves, a pure function of the reconstructed domain like the vanishing table. `from_slice` checks only its domain and scalar canonicity, so a modified nonzero table passes checked deserialization and surfaces later as proofs that fail verification — the failure mode #902 describes.

**Fix:** File a follow-up issue mirroring #902 for `linear_evaluations`; the same recurrence-based validation applies (compare against `GENERATOR * group_gen^i`).

moCello (approved):
APPROVE — the validation is exact: the accept set shrinks to precisely the compile-produced table, subsuming the old zero check with the wire format and valid-artifact round-trips intact, and both regressions from #902 plus the round-trip controls are in place. Two notes below, neither gating, and they have different homes. The predicate hardening is the only one in scope here: take it on this branch in one push, or track it in an issue — not an untracked later change. The sibling-field gap is pre-existing and wants its own issue regardless of what happens with the first.

PR #905, opened 2026-08-07
Validate commitment keys in RKYV public parameters

### LINKED ISSUE #903: Validate commitment keys in RKYV public parameters
## Summary

Checked RKYV deserialization of `CommitKey` and `PublicParameters` validates the archive structure, but currently accepts an empty commitment key and malformed raw `G1Affine` representations.

The derived point validation only checks the archived fields structurally. It does not require the raw field limbs to be canonical or the archived infinity byte to be zero or one. An invalid infinity byte can therefore panic during checked decoding in debug builds, while an equivalent non-canonical field representation can be accepted as a live point outside the guarantees of `G1Affine`.

An empty key also reaches `CommitKey::max_degree` and `truncate`, where it panics with overflow checks enabled or after the wrapped degree calculation with overflow checks disabled.

## Expected behavior

- RKYV commitment-key points are stored and decoded through their canonical compressed representation.
- Checked RKYV deserialization rejects empty commitment keys.
- Malformed point encodings return an error without panicking in debug or release builds.
- Only canonical, on-curve points in the correct subgroup reach a live `CommitKey` through checked RKYV decoding.
- Valid commitment keys and public parameters continue to round-trip and compile circuits normally.

## Validation

- Add raw-byte regressions for an invalid infinity byte and a non-canonical field coordinate.
- Add malformed compressed-point and empty-key regressions.
- Run the malformed raw archive cases with debug assertions enabled and disabled.
- Keep valid `CommitKey` and `PublicParameters` RKYV round-trips and a circuit compilation control.

HDauven:
## Summary

- archive commitment-key points as canonical compressed encodings
- reject empty, malformed, off-curve, and non-prime-order commitment keys before constructing live points
- keep semantic point validation parallel under `std` and sequential under `no_std`
- cover direct `CommitKey` and complete `PublicParameters` round-trips, malformed raw archives, and circuit compilation

## Validation

- `make fmt`
- `make clippy`
- `make no-std`
- `cargo check --lib --no-default-features --features alloc,rkyv-impl,rkyv/size_32`
- `cargo clippy --lib --features rkyv-impl,rkyv/size_32 -- -D warnings`
- `cargo doc --no-deps --features rkyv-impl,rkyv/size_32`
- `make test`

Checked RKYV decode benchmark (release, 262,144 G1 points): approximately 2.25 s and a 12,583,160-byte archive. The first draft's raw-point validation took approximately 1.85 s and produced a 27,263,224-byte archive; canonical encoding therefore adds about 0.4 s while reducing archive size by roughly 54%. Master took approximately 9 ms because it performed no semantic point validation.

This is a checked parameter-load cost only. Circuit gates, proving, and verification are unchanged.

Resolves #903

### DIFF
--- a/src/commitment_scheme/kzg10/key.rs
+++ b/src/commitment_scheme/kzg10/key.rs
@@ -19,6 +19,8 @@ use merlin::Transcript;
 use rkyv::{
     Archive, Deserialize, Fallible, Serialize,
     ser::{ScratchSpace, Serializer},
+    validation::ArchiveContext,
+    vec::ArchivedVec,
     with::{ArchiveWith, DeserializeWith, SerializeWith},
 };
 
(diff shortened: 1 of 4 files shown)


src/commitment_scheme/kzg10/srs.rs
```diff
@@ -317,6 +363,87 @@ mod test {
...
+    #[test]
+    fn rkyv_rejects_raw_archive_with_invalid_infinity_flag() {
+        let pp = PublicParameters::setup(8, &mut OsRng).unwrap();
+        let mut bytes =
+            rkyv::to_bytes::<_, 256>(&RawPublicParameters::from(&pp)).unwrap();
+        let point = first_raw_commitment_point_offset(&bytes);
+        bytes[point + 96] = 2;
+
+        assert_archived_parameters_rejected_without_panic(&bytes);
+    }
```
moCello: **NOTE:** Both raw-archive tests pass independently of their mutations — a `RawPublicParameters` archive is rejected as a whole, so the properties in the test names are never isolated.

**Context:** `RawPublicParameters` archives 104-byte raw points while the new decoder expects 48-byte compressed elements, so the commit-key region is invalid with overwhelming probability before any mutation. The tests still guard one real property — legacy-layout archives are rejected without panicking. The named mutations are isolated on the new format by `commit_key_rkyv_round_trip_rejects_malformed_points` and `rkyv_rejects_invalid_archived_commit_key_points`.

**Fix:** Rename the two tests to state what they verify — panic-free rejection of legacy raw-layout archives — or fold them into one legacy-rejection regression next to `rkyv_archive_omits_prepared_points_and_rejects_legacy_layout`.

src/commitment_scheme/kzg10/key.rs
```diff
@@ -32,19 +34,165 @@ use crate::util;
...
+        &self,
+        pos: usize,
+        resolver: Self::Resolver,
+        out: *mut Self::Archived,
+    ) {
+        unsafe {
+            CommitKeyArchive::from(self).resolve(pos, resolver, out);
+        }
+    }
+}
```
moCello, line 161: **NOTE:** Serialization builds the full `CommitKeyArchive` twice — in `serialize` and again in `resolve` — though `resolve` only needs the vector length.

**Impact:** For a production-size key (2^18 points) the second conversion is a ~12 MB temporary and a full canonical-encoding pass over every point, on the archive-writing path only; checked decode is unaffected.

**Fix:** Resolve the archived vec from `self.powers_of_g.len()` via `ArchivedVec::resolve_from_len` instead of converting the key a second time; `ArchivedVec::serialize_from_iter` on the serialize side would additionally drop the intermediate `Vec`. Both APIs exist in rkyv 0.7 and are drop-in.

moCello (approved):
APPROVE — the change delivers everything #903 asks for: the archive stores canonical compressed points, checked decoding rejects empty, malformed, off-curve, and non-prime-order keys without panicking, the validation split between `CheckBytes` and `Deserialize` is cryptographically sound, and the design matches the merged opening-key precedent. Two notes below, both small; neither gates the merge. If you take them, batch both into a single push — as many commits as the changes warrant — so the approval re-opens once; otherwise they belong in a tracked issue rather than an untracked follow-up.

src/commitment_scheme/kzg10/srs.rs
```diff
@@ -317,6 +363,47 @@ mod test {
...
+
+    #[cfg(feature = "rkyv-impl")]
+    #[test]
+    fn rkyv_rejects_legacy_raw_commitment_key_layout_without_panicking() {
+        let pp = PublicParameters::setup(8, &mut OsRng).unwrap();
+        let bytes =
+            rkyv::to_bytes::<_, 256>(&RawPublicParameters::from(&pp)).unwrap();
+
+        assert_archived_parameters_rejected_without_panic(&bytes);
```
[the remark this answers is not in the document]
HDauven: Those mutations would not isolate the historical representations here. `RawPublicParameters` stores 104-byte raw points, while the new decoder expects 48-byte compressed elements, so the legacy archive is rejected at the layout boundary independently of either mutation. That is why Moana asked us to consolidate the two tests into this accurately named panic-free legacy-layout regression. Malformed flags, non-canonical coordinates, off-curve points, and subgroup failures are isolated separately against the current compressed format.

PR #907, opened 2026-08-10
Prevent malformed point witnesses from panicking

### LINKED ISSUE #906: Make point-arithmetic witness generation panic-free
## Summary

Point membership constraints can reject malformed witness coordinates, but the native witness-generation path still performs unchecked group arithmetic before those constraints are evaluated.

`add_point_gates` constructs affine points from witness coordinates, adds them natively, and converts the extended result back to affine. A crafted off-curve input can make that intermediate result have `Z = 0`, causing affine conversion to panic at `invert().unwrap()` instead of letting proving return an unsatisfied-constraint error.

This is a prover-availability issue, not a proof-soundness bypass.

## Expected behavior

- Native point-arithmetic witness generation remains total for malformed assignments that constraints will reject.
- Invalid point witnesses return a proving error without panicking.
- Honest point addition and scalar multiplication produce the same witnesses and circuit shape as before.

## Validation

- Add a regression using the crafted off-curve point and scalar multiplication path that currently produces an intermediate `Z = 0`.
- Require proving to return an error without unwinding.
- Keep honest point-addition and scalar-multiplication controls.
- Confirm the regression fails again when the witness-generation guard is removed.

HDauven:
## Summary

- keep variable-base point witness generation total when an intermediate extended sum has no affine image
- preserve the circuit shape and let the existing constraints reject malformed assignments
- cover the denominator-zero regression with an honest proving control

## Validation

- `cargo test --release --lib composer::tests::soundness::point::mul_point_rejects_malformed_base_without_panicking -- --exact --nocapture`
- `cargo test --release --lib composer::tests::soundness::point::mul_point_layout_matches_golden -- --exact`
- `make fmt`
- `make clippy`
- `make no-std`
- `make test`

Resolves #906

### DIFF
--- a/src/composer/tests/soundness/point.rs
+++ b/src/composer/tests/soundness/point.rs
@@ -342,6 +342,69 @@ fn torsion_free_layout_matches_golden() {
     );
 }
 
+struct MulPointCircuit {
+    scalar: JubJubScalar,
+    point: JubJubAffine,
+}
+
+impl Default for MulPointCircuit {
+    fn default() -> Self {
+        Self {
+            scalar: JubJubScalar::from(3u64),
+            point: prime_order_point(),
+        }
+    }
+}
+
+impl Circuit for MulPointCircuit {
+    fn circuit(&self, composer: &mut Composer) -> Result<(), Error> {
+        let scalar = composer.append_witness(self.scalar);
+        let point = composer.append_point(self.point)?;
+        let point = composer.assert_torsion_free_point(point);
+        composer.component_mul_point(scalar, point);
+
+        Ok(())
+    }
+}
+
+#[test]
+fn mul_point_rejects_malformed_base_without_panicking() {
+    let mut rng = StdRng::seed_from_u64(0x906);
+    let pp = PublicParameters::setup(1 << 11, &mut rng).expect("setup");
+    let (prover, verifier) =
+        Compiler::compile::<MulPointCircuit>(&pp, b"mul-point-malformed-base")
+            .expect("compile");
+
+    let accepted = MulPointCircuit::default();
+    assert_verifies(&prover, &verifier, &mut rng, &accepted);
+
+    let x = BlsScalar::from_raw([
+        0x3218_5d43_5879_5954,
+        0x018f_1597_d916_ddf5,
+        0xf49c_53d3_e92b_3582,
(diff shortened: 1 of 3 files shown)


src/composer/tests/soundness/point.rs
```diff
@@ -342,6 +342,69 @@ fn torsion_free_layout_matches_golden() {
...
+        scalar: JubJubScalar::from(3u64),
+        point: JubJubAffine::from_raw_unchecked(x, BlsScalar::one()),
+    };
+
+    // Pin the fixture to the affine-projection failure: doubling remains
+    // projectable, while the final 2P + P addition has no affine image.
+    let point = JubJubExtended::from(rejected.point);
+    let doubled = point + rejected.point;
+    assert_ne!(doubled.get_z(), BlsScalar::zero());
+    assert_eq!((doubled + rejected.point).get_z(), BlsScalar::zero());
```
moCello, line 397: **NOTE:** Follow-up, non-blocking: the module doc's "No case below exercises either" pole claim is now stale, and the guard's pole-rejection argument stays untested in isolation.

**Context:** A vanishing extended `Z` is exactly the helper wire at a pole `±1/(D·y_1·x_2)`, so this fixture constructs what the module doc (src/composer/tests/soundness/point.rs:22-32) says no case below constructs — and via off-curve addends, a case outside both prongs of that doc's argument. The identity fallback in `add_point_gates` rests on the pole gate collapsing to a provably nonzero numerator (`d` non-square, `−1` a square), and this test cannot pin that: `assert_torsion_free_point` runs first on the off-curve base, so its gates already return `CircuitUnsatisfied` even if the pole gate were satisfiable.

**Fix:** In a follow-up change, not this PR: update the module doc paragraph for the new fixture, and add an isolated regression — emit only the addition gates for a Z-vanishing pair (via the `add_point_gates` seam, as the existing forgery mirrors do) so the pole gate is the sole rejecting constraint.

moCello (approved):
APPROVE — The guard is sound: a vanished addition denominator leaves its gate check unsatisfiable for any substituted witness, the regression fails by the original panic without the guard, and the layout-equality and golden-digest tests hold the circuit shape unchanged. The single note is follow-up material and asks nothing of this branch.

xevisalle (approved):
lgtm

PR #909, opened 2026-08-10
Validate cached linear evaluations

### LINKED ISSUE #908: Validate cached linear evaluations when decoding Provers
## Summary

`ProverKey::from_slice` accepts serialized `perm_linear_evaluations` after checking their encoding, length, and evaluation domain.

The compiler derives this table as `domain_8n.coset_fft(&[BlsScalar::zero(), BlsScalar::one()])`, so every entry is determined by the reconstructed domain. A modified but canonically encoded nonzero table can therefore pass checked decoding and only fail later during proof generation or verification.

## Expected behavior

- Checked `ProverKey` deserialization verifies the cached linear evaluations against values derived from the validated domain, or reconstructs the table instead of trusting it.
- Any modified cached evaluation is rejected at the artifact boundary.
- Valid serialized Provers continue to round-trip, prove, and verify normally.

## Validation

- Add a regression that changes one cached evaluation to another nonzero scalar and requires `ProverKey` deserialization to reject it.
- Mirror the corruption through complete `Prover` decoding.
- Keep a valid Prover round-trip and proving control.

HDauven:
Resolves #908

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Validate cached permutation linear evaluations during checked `Prover`
+  deserialization [#908]
 - Validate cached vanishing-coset evaluations during checked `Prover`
   deserialization [#902]
 - Validate nonempty, canonically encoded, on-curve, prime-order KZG commitment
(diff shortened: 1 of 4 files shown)


moCello (approved):
APPROVE — the validation matches the closed form the compiler produces (`coset_fft(&[0,1])` is `GENERATOR * group_gen^i` in natural order), runs against the canonically reconstructed domain rather than attacker-supplied bytes, and the regression tests cover every path the issue asked for: ProverKey-level rejection, full Prover-decode rejection, and closed-form agreement with length and mismatch negatives. The change is minimal, mirrors the established vanishing-evaluation check, and the changelog and issue linkage are in order.

Issue #861, opened 2026-01-25
Move to Rust edition 2024


HDauven:
#### Summary

Upgrade the workspace from Rust edition 2021 to edition 2024 to stay current with the ecosystem, benefit from improved language defaults/lints, and reduce long-term maintenance. Expected outcome: the repo compiles and passes CI under edition 2024 with stable Rust.

#### Possible solution design or implementation

We can run `cargo fix --edition` and resolve any follow-up warnings or errors.

We need to set the workspace resolver from `2` to `3`.

Issue #586, opened 2021-09-27
Plonkup - Add test coverage for polynomial evaluations


vlopes11:
**Describe what you want implemented** 
Currently the new polynomial evaluations for plonkup aren't tested

**Describe "Why" this is needed**
It is critical to have edge cases and test vectors covered in order to have a robust implementation

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #912, opened 2026-08-10
Test curve-addition pole rejection in isolation

### LINKED ISSUE #910: Isolate curve-addition pole rejection in point soundness tests
## Summary

The point-soundness module documentation says that no test constructs a curve-addition pole. The malformed scalar-multiplication fixture from #906 now does construct off-curve addends whose extended sum has `Z = 0`, so that statement is stale.

The existing end-to-end regression also applies point-membership and torsion constraints before the relevant addition. Those constraints already reject the malformed base, so the test does not isolate the curve-addition pole gate that rejects the fallback witness.

## Expected behavior

- The module documentation accurately describes the pole coverage and the off-curve #906 fixture.
- A regression uses the private `add_point_gates` seam to emit only the addition gates for a `Z`-vanishing pair.
- The curve-addition pole constraint is the sole rejecting constraint for that pair.
- Production circuit behavior and layout remain unchanged.

## Validation

- Add an honest addition control through the same private seam.
- Require the isolated `Z`-vanishing assignment to return `CircuitUnsatisfied`.
- Include a negative control demonstrating that no unrelated constraint causes the rejection, if practical.

HDauven:
## Summary

- update the point-soundness documentation for the `Z = 0` pole fixture
- exercise the private `add_point_gates` seam without membership constraints
- prove the same malformed witnesses when the sole curve-addition selector is disabled
- leave production circuit behavior and layout unchanged

## Validation

- `cargo test --release --lib composer::tests::soundness::point::curve_addition_pole_rejected_in_isolation -- --exact --nocapture`
- `cargo test --release --lib composer::tests::soundness::point::mul_point_rejects_malformed_base_without_panicking -- --exact --nocapture`
- `cargo test --release --lib composer::tests::soundness::point::component_add_point_layout_matches_golden -- --exact`
- `make test`
- `make clippy`
- `make no-std`
- `make fmt`
- `git diff --check`

Resolves #910

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `Error::CircuitUnsatisfied` [#877]
 - Add `Display` for `Error` to builds without the `std` feature [#895]
 - Add `core::error::Error` for `Error` to builds without the `std` feature [#895]
+- Cover curve-addition poles directly in point soundness regressions and update
+  their module documentation [#910]
 
 ### Changed
 
@@ -750,6 +752,7 @@ is necessary since `rkyv/validation` was required as a bound.
 
 <!-- ISSUES -->
 [#913]: https://github.com/dusk-network/plonk/issues/913
+[#910]: https://github.com/dusk-network/plonk/issues/910
 [#908]: https://github.com/dusk-network/plonk/issues/908
 [#902]: https://github.com/dusk-network/plonk/issues/902
 [#903]: https://github.com/dusk-network/plonk/issues/903
--- a/src/composer/tests/soundness/point.rs
+++ b/src/composer/tests/soundness/point.rs
@@ -23,13 +23,18 @@
 //! for a matching `(x_3, y_3)` at every wire value but the two poles
 //! `±1/(D*y_1*x_2)`, where one check degenerates: at `+` the `y_3` check
 //! collapses to `y_1*y_2 + x_1*x_2 = 0` and leaves `y_3` free, at `-` the `x_3`
-//! check collapses to `x1_y2 + y1_x2 = 0` and leaves `x_3` free. **No case
-//! below exercises either** — `forced_output` inverts both denominators, so it
-//! panics on a pole rather than constructing one. Neither is an escape hatch,
-//! but that is an argument rather than a test: with `a = -1` a square and `d` a
-//! non-square (asserted in [`subgroup_parameters_hold`]) the addition law is
-//! complete, so no on-curve addends put the *honest* wire on a pole, and a
-//! forged pole wire still carries a nonzero binding residual.
+//! check collapses to `x1_y2 + y1_x2 = 0` and leaves `x_3` free. The
+//! helper-wire forgeries below avoid both poles — `forced_output` inverts both
+//! denominators — but the malformed scalar-multiplication fixture constructs
+//! off-curve addends whose extended sum has `Z = 0`. The isolated pole
+//! regression emits only `add_point_gates` for that pair; disabling its sole
+//! curve-addition selector makes the same assignment prove, pinning rejection
+//! to the pole gate. Neither pole is an escape hatch: with `a = -1` a square
+//! and `d` a non-square (asserted in [`subgroup_parameters_hold`]) the addition
+//! law is complete, so no on-curve addends put the *honest* wire on a pole. For
+//! off-curve addends at an honest-wire pole, the collapsed output numerator
+//! cannot also vanish without making `d` a square; a helper wire forged to a
+//! pole instead carries a nonzero binding residual.
 //!
 //! So without that term a prover picks the wire freely and still satisfies
 //! every other constraint — including, as the steering case shows, picking it
@@ -52,7 +57,8 @@ use rand::SeedableRng;
 use rand::rngs::StdRng;
 
 use super::support::{
-    assert_rejected, assert_verifies, gate_digest, invert, steering_target,
+    assert_rejected, assert_verifies, gate_digest, gate_layout, invert,
+    steering_target,
 };
 use crate::composer::point::EIGHT_INV;
 use crate::composer::{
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -66,6 +66,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Reject trailing data when deserializing compressed circuits [#913]
+- Cover curve-addition poles directly in point soundness regressions and update
+  their module documentation [#910]
```
moCello: **NOTE:** The entry sits under `Fixed`, but nothing was fixed — `MultisigNonce`-style behaviour is unchanged and the release is identical with or without this PR. Test-coverage entries in this file live under `Added`.

**Context:** The change is confined to an inline `#[cfg(test)]` module and its doc, so nothing reaches the crate surface; rustdoc does not publish it. `CONTRIBUTING.md:41-44` asks contributors to "avoid adding things like doc-nitpicks and similar changes which do not affect directly any added, fixed, removed or changed feature", which read literally would drop the entry — but the file's own practice is plainly otherwise, so the heading is the live question rather than the entry's existence. `CHANGELOG.md:401` records "Add test coverage for polynomial evaluations [#586]" under `Added`, `:122` "Added more tests for low coverage parts of the code [#861]", and `:293-301` nine consecutive "Add and restructure tests for …" entries.

**Fix:** Move the entry under `Added`, keeping the `[#910]` link definition. `:401` is the closest precedent for its wording.

src/composer/tests/soundness/point.rs
```diff
@@ -405,6 +423,102 @@ fn mul_point_rejects_malformed_base_without_panicking() {
...
+    >(&pp, b"isolated-curve-addition-pole")
+    .expect("compile isolated curve addition");
+    assert_verifies(&prover, &verifier, &mut rng, &accepted);
+    assert_rejected(
+        &prover,
+        &mut rng,
+        &accepted,
+        &rejected,
+        "Z-vanishing curve addition",
+    );
```
moCello, line 507: **NOTE:** Follow-up rather than work for this PR — the rejection is overdetermined, so what the fixture pins is the gate, not the pole.

**Context:** The pair hits the `d·x1_y2·y1_x2 = -1` pole, where the `x_3` residual collapses to `x1_y2 + y1_x2`: nonzero, and independent of whatever `x_3` is assigned. But the `Z = 0` fallback also assigns `y_3 = 1` against a `y_3` check whose coefficient is `2` here and still binding, so that residual is nonzero as well. A witness that merely got `y_3` wrong is turned away identically.

Pinning the gate is what #910 asked for, and the selector-disabled control delivers it. What stays untested is the next step in the module doc's argument — that a prover cannot exploit the coordinate the pole frees — on which the rewrite drops the previous text's "that is an argument rather than a test" caveat.

**Fix:** Emit the same gate pair with an honest `x_1·y_2` wire, `x_3` from `steering_target()`, and `y_3` solved from the still-binding output check — `(y_1y_2 + x_1x_2)·(1 − d·x1_y2·y1_x2)⁻¹`, which at this pole is `(y_1y_2 + x_1x_2)/2`, **not** the `(0, 1)` fallback's `y_3 = 1`. Required to be `CircuitUnsatisfied`; the collapsed numerator is then the only nonzero residual. `forged_double` is the model, generalised to distinct addends — but note `forced_output` cannot be reused here, since it inverts `1 + skew`, which is zero at this pole. Cheaper complement: an off-curve pair away from a pole, required to verify — the gate identity references no curve equation, so off-curveness alone does not reject.

src/composer/tests/soundness/point.rs
```diff
@@ -405,6 +423,102 @@ fn mul_point_rejects_malformed_base_without_panicking() {
...
+        let b = composer.append_point(self.b)?;
+        let addition_row = composer.constraints.len();
+        composer.add_point_gates(a, b);
+
+        if !ENABLED {
+            let selector =
+                &mut composer.constraints[addition_row].q_variable_group_add;
+            assert_eq!(*selector, BlsScalar::one());
+            *selector = BlsScalar::zero();
+        }
```
moCello, line 464: **NOTE:** The debugger keeps the pre-mutation row, so the negative control's diagnostic names a constraint the compiled circuit does not have.

**Context:** `append_custom_gate` emits `ConstraintAppended` with the constraint as built — before this block runs — and `Debugger` stores that copy. `ProofFinished` then calls `report_unsatisfied` unconditionally, so under `--all-features`, which `make test` runs, the debugger evaluates a `q_variable_group_add = 1` row against the `Z = 0` fallback and reports it unsatisfied while the proof succeeds. That surfaces under `--nocapture` and in the captured output of any future failure here, where `unsatisfied_report`'s own doc reads the pattern as a constraint missing from the verifier key — the one conclusion that is wrong.

**Fix:** Note the stale debugger copy in the comment on this block so the diagnostic is not chased later. Hand-emitting the row would avoid it but duplicates the seam's `Z = 0` fallback in the test; the mutation is the better trade, it just needs saying.

moCello (approved):
APPROVE — the isolation holds. With `Z = 1` the extended addition projects to exactly the affine rational formula the widget checks, using no curve equation at any step, and `Z_3 = 4(1 - d²P²)` vanishes precisely at the honest helper-wire poles — so the fixture's `get_z() == 0` assertion is the pole condition rather than a proxy for it. The doc's argument that the collapsed numerator cannot also vanish is correct at both poles; at the `+` pole it leans on `-1` being a square and `d` not, both of which `subgroup_parameters_hold` already asserts. Production layout is untouched, so no consumer verifier key moves. Nothing gates the merge.

Three notes below, none blocking. Two are small enough to ride any further push: the changelog entry sits under `Fixed` when nothing was fixed, and the negative control's stale debugger row deserves a comment so the diagnostic is not chased later.

The third is follow-up material rather than work for this PR: the rejected fixture pins the curve-addition gate — which is what #910 asked for — rather than the pole itself, since at this pole the `y_3` residual is nonzero too and a witness that merely got `y_3` wrong is turned away identically. Closing that means a new fixture, not an edit to this one.

PR #915, opened 2026-08-11
Isolate collapsed curve-addition pole residual

### LINKED ISSUE #914: Isolate the collapsed residual at a curve-addition pole
## Summary

The point-soundness regression added for #910 establishes that a malformed assignment at a curve-addition pole is rejected by the curve-addition gate. Its selector-disabled control also proves that no unrelated gate rejects the assignment.

The rejection is still overdetermined. At the exercised pole, the x-coordinate output check collapses and leaves the assigned x-coordinate free, but the fallback y-coordinate also violates the remaining binding output check. The regression therefore pins the complete gate rather than demonstrating that the collapsed numerator alone prevents an escape.

## Expected behavior

- Construct off-curve addends at an honest-helper-wire pole.
- Assign the helper wire to the actual x₁ · y₂ product.
- Assign the coordinate freed by the pole independently.
- Solve the remaining binding output equation instead of using the affine-projection fallback.
- Require proving to return Error::CircuitUnsatisfied because the collapsed numerator remains nonzero.
- Keep production circuit behavior and layout unchanged.

## Validation

- Pin the selected pole condition explicitly.
- Confirm that every residual except the collapsed numerator is satisfied by the constructed witness.
- Add an off-curve pair away from a pole as a positive control if needed to demonstrate that off-curveness alone is not rejected by the raw addition gate.
- Preserve the existing isolated gate and selector-disabled controls from #910.

HDauven:
Resolves #914

### DIFF
--- a/src/composer/tests/soundness/point.rs
+++ b/src/composer/tests/soundness/point.rs
@@ -27,14 +27,17 @@
 //! helper-wire forgeries below avoid both poles — `forced_output` inverts both
 //! denominators — but the malformed scalar-multiplication fixture constructs
 //! off-curve addends whose extended sum has `Z = 0`. The isolated pole
-//! regression emits only `add_point_gates` for that pair; disabling its sole
-//! curve-addition selector makes the same assignment prove, pinning rejection
-//! to the pole gate. Neither pole is an escape hatch: with `a = -1` a square
-//! and `d` a non-square (asserted in [`subgroup_parameters_hold`]) the addition
-//! law is complete, so no on-curve addends put the *honest* wire on a pole. For
-//! off-curve addends at an honest-wire pole, the collapsed output numerator
-//! cannot also vanish without making `d` a square; a helper wire forged to a
-//! pole instead carries a nonzero binding residual.
+//! regressions emit only `add_point_gates` for that pair; disabling its sole
+//! curve-addition selector makes the malformed fallback assignment prove,
+//! pinning rejection to the pole gate. A focused witness keeps the helper wire
+//! honest, chooses the x-coordinate the pole frees independently, and solves
+//! the still-binding y-coordinate check, leaving only the collapsed nonzero
+//! x-coordinate numerator. Neither pole is an escape hatch: with `a = -1` a
+//! square and `d` a non-square (asserted in [`subgroup_parameters_hold`]) the
+//! addition law is complete, so no on-curve addends put the *honest* wire on a
+//! pole. For off-curve addends at an honest-wire pole, the collapsed output
+//! numerator cannot also vanish without making `d` a square; a helper wire
+//! forged to a pole instead carries a nonzero binding residual.
 //!
 //! So without that term a prover picks the wire freely and still satisfies
 //! every other constraint — including, as the steering case shows, picking it
@@ -42,11 +45,12 @@
 //! the attacker's choosing. Those forgeries reproduce the production emission
 //! with only the wire falsified and the coordinates its own checks force;
 //! `forced_output` does that solving, so it is itself pinned against
-//! `compute_quotient_i`. That pin is this module's only reach into
-//! `proof_system::widget`, and is deliberate: it exists to tie this module's
-//! `forced_output` to the widget's own identity, and building the `ProverKey`
-//! by struct literal makes a new field on it break the build here rather than
-//! silently weaken the pin.
+//! `compute_quotient_i`. The focused pole fixture separately pins its collapsed
+//! residual against the same production identity. Those pins are this module's
+//! only reaches into `proof_system::widget`, and are deliberate: they tie the
+//! local residual constructions to the widget's own identity, while building
+//! the `ProverKey` by struct literal makes a new field on it break the build
+//! here rather than silently weaken either pin.
 
 use dusk_bls12_381::BlsScalar;
 use dusk_jubjub::{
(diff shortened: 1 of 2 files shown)


src/composer/tests/soundness/point.rs
```diff
@@ -521,6 +610,178 @@ fn curve_addition_pole_rejected_in_isolation() {
...
+        &output.get_v(),
+        &x_2,
+        &y_2,
+        &x1_y2,
+    );
+    assert_eq!(
+        widget_identity,
+        residuals[1] * challenge.square() * challenge,
+        "the production identity must contain only the collapsed x3 residual",
+    );
```
moCello: **NOTE:** The x3 collapse is demonstrated only against the local residual copy — one equality at one challenge is the whole tie to `compute_quotient_i`.

**Context:** The freedom assertion at point.rs:690-698 re-runs `curve_addition_residuals` on a perturbed x3, reading the collapse off the transcription rather than off the widget. With the selector evaluation fixed at `1` the identity is `(r0 + r1·κ + r2·κ²)·c` with `κ = c²`, and the honest wire makes `r0` vanish identically, so this equality fixes `r1 + r2·c²` — one linear combination, not either residual. The doc block at point.rs:1233-1236 makes that argument for the sibling pin and closes it with two challenges of distinct squares.

**Severity-rationale:** Nothing escapes today: `forced_output_satisfies_the_widget_output_checks` catches redistribution between the widget's output terms, and the off-curve control's `[zero; 3]` catches drift in the first or third transcribed residual.

**Fix:** Evaluate the identity at the perturbed x3 too, and at a second challenge with a distinct square — the collapse and the y3 binding then read off the production identity directly.

src/composer/tests/soundness/point.rs
```diff
@@ -521,6 +610,178 @@ fn curve_addition_pole_rejected_in_isolation() {
...
+    assert_verifies(&prover, &verifier, &mut rng, &off_curve_control);
+
+    assert_rejected(
+        &prover,
+        &mut rng,
+        &accepted,
+        &rejected,
+        "collapsed curve-addition pole residual",
+    );
```
moCello, line 824: **NOTE:** The focused fixture has no selector-disabled control, so its rejection is attributed to the curve-addition gate by construction rather than demonstrated.

**Context:** `Error::CircuitUnsatisfied` names no gate — it comes from one degree check over the whole quotient polynomial (src/proof_system/quotient_poly.rs:131), and `curve_addition_pole_rejected_in_isolation` supplies the attribution by pairing its rejection with `IsolatedCurveAdditionCircuit::<false>::pole()` proving. The curve-addition row is the only active constraint over these witnesses today, so the attribution holds; what the control adds is a guard against a constraint later added to the seam or to `append_point` rejecting this fixture for a new reason with the test still green. `pole_with_solved_y3` is already generic over `ENABLED`, so `<false>::pole_with_solved_y3()` is constructible today.

**Fix:** Compile the `<false>` circuit here too and `assert_verifies` the same assignment with the curve-addition selector cleared.

src/composer/tests/soundness/point.rs
```diff
@@ -985,15 +1238,19 @@ fn curve_addition_binds_the_helper_wire() {
 /// That equality would also hold for a widget with both output checks deleted,
 /// the identity then being the binding term outright, so each coordinate is
 /// perturbed and required to move the identity off it.
-#[test]
-fn forced_output_satisfies_the_widget_output_checks() {
+fn curve_addition_prover_key() -> curve_addition::ProverKey {
```
moCello: **NOTE:** Two docs still describe the pre-refactor shape: the extracted factory inherited the test's rationale block, and the module doc calls the widget pin the module's only one.

**Context:** point.rs:1228-1240 describes `forced_output_satisfies_the_widget_output_checks`' assertion strategy, not what the factory does — that test is now undocumented, and the two-challenge reasoning no longer sits beside the code implementing it. point.rs:47-49 reads "That pin is this module's only reach into `proof_system::widget`", which the new pin on `curve_addition_residuals` makes a second; that paragraph is where a later reader checks before adding one.

**Fix:** Move the block back onto `forced_output_satisfies_the_widget_output_checks`, leave `curve_addition_prover_key` a one-line doc for what it is, and widen the module-doc sentence to cover both pins.

src/composer/tests/soundness/point.rs
```diff
@@ -27,14 +27,17 @@
...
-//! pole instead carries a nonzero binding residual.
+//! regressions emit only `add_point_gates` for that pair; disabling its sole
+//! curve-addition selector makes the malformed fallback assignment prove,
+//! pinning rejection to the pole gate. A focused witness keeps the helper wire
+//! honest, chooses the x-coordinate the pole frees independently, and solves
+//! the still-binding y-coordinate check, leaving only the collapsed nonzero
+//! x-coordinate numerator. Neither pole is an escape hatch: with `a = -1` a
+//! square and `d` a non-square (asserted in [`subgroup_parameters_hold`]) the
+//! addition law is complete, so no on-curve addends put the *honest* wire on a
+//! pole. For off-curve addends at an honest-wire pole, the collapsed output
```
moCello, line 38: **NOTE:** The `+` pole has no regression anywhere — every fixture sits at `skew = -1`, so half of "neither pole is an escape hatch" stays prose-only. Follow-up, not this PR.

**Context:** `z_vanishing_addends` is the module's only pole fixture and this change pins it to `skew = -1`. The `+` pole — `skew = 1`, where `1 - skew` zeroes the y3 coefficient and leaves `y_1·y_2 + x_1·x_2` as the residual — is constructible off-curve by fixing `x_1, y_1, x_2` nonzero and solving `y_2 = 1/(D·x_1·x_2·y_1)`; the argument that its residual cannot vanish mirrors this change's, since `y_1y_2 + x_1x_2 = 0` there gives `d = -(1/x_1x_2)²`, a square because `-1` is one.

**Severity-rationale:** The gap predates this change — #914 scopes itself to the exercised pole and the diff delivers it.

**Fix:** Track a follow-up for a `+`-pole fixture mirroring `pole_with_solved_y3` — honest helper wire, y3 chosen independently, x3 solved through the still-binding check.

moCello (approved):
APPROVE — The construction does what #914 asked: the helper wire stays honest, x3 is chosen independently, y3 solves the still-binding check, and the collapsed numerator is the only nonzero residual left, with the shared seam emitting `add_point_gates`' rows witness-for-witness so the layout comparison in `assert_rejected` keeps its teeth. The transcribed residuals match the widget sign for sign, skew formula included, and nothing outside the tests changes.

Four notes below, none blocking and all in test code: the pin at the pole reads the collapse off the local residual copy rather than off `compute_quotient_i`, the focused fixture has no selector-disabled control, and the refactor left two docs describing the pre-refactor shape. Yours to take or leave, and worth batching into one push if you take any.

The last is a follow-up rather than anything for this branch: with the `-` pole now argued end to end, the `+` pole is the module's only remaining prose-only completeness claim — no fixture anywhere sits at `skew = 1`.

src/composer/tests/soundness/point.rs
HDauven: Done. The focused and perturbed x3 assignments now run through compute_quotient_i at two challenges with distinct squares.

src/composer/tests/soundness/point.rs
HDauven: Done. The exact focused assignment now verifies in a separately compiled selector-disabled circuit.

src/composer/tests/soundness/point.rs
HDauven: Done. The rationale now documents the test it supports, and the module documentation describes both production-widget pins.

src/composer/tests/soundness/point.rs
HDauven: Tracked separately in #916.

PR #922, opened 2026-08-14
Group verifier MSM terms

### LINKED ISSUE #920: Group verifier MSM terms into the final pairing input
## Current behavior

The PLONK verifier currently materializes the linearization commitment MSM and the batched-opening commitment MSM separately, performs the generator and opening-witness scalar multiplications independently, and normalizes the two final G1 pairing inputs separately. These independent operations add avoidable verification overhead.

## Proposed improvement

Append the unchanged linearization, batched-opening, evaluation-generator, and opening-witness scalar/base pairs into one variable-base MSM for the right pairing input. Keep the left pairing equation unchanged, then batch-normalize the two final projective pairing inputs together.

## Acceptance criteria

- Preserve the current and legacy verifier equations, scalar signs and point pairings exactly.
- Do not change Fiat–Shamir transcript construction, commitment/evaluation semantics, pairing order, errors, public API, feature gates, or `no_std` support.
- Add focused equivalence coverage for the regrouped equation and retain successful proof verification.
- Demonstrate a statistically meaningful improvement with the existing end-to-end verifier benchmark.
- Pass release tests with default and all features, `no_std` builds, formatting, and Clippy.

HDauven:
Closes #920.

## Summary

- fold the current and legacy verifier linearization, batched-opening, evaluation-generator, and opening-witness terms into one variable-base MSM for the right pairing input
- batch-normalize the unchanged left pairing input and regrouped right pairing input together
- retain the current v2/v3 selector commitments and the legacy commitment list exactly
- add test-only reconstruction of the previous separate right-input equation for direct equivalence checking

The transcript, challenges, scalar/commitment ordering, error behavior, public API, feature gates, and G2 pairing-input order are unchanged.

## Benchmark

Existing end-to-end Criterion verifier benchmark:

```text
RAYON_NUM_THREADS=22 cargo bench --bench plonk -- 'Verify' --save-baseline verifier-msm-base
RAYON_NUM_THREADS=22 cargo bench --bench plonk -- 'Verify' --baseline verifier-msm-base
```

Environment: Rust 1.94.0, WSL2, Intel Core Ultra 9 275HX, 22 logical CPUs, Criterion sample size 10 with a 3-second warm-up. Each benchmark case generated and successfully sanity-verified a real proof before timing verification. The host had variable background load, so the complete per-size medians are included:

| Gates | Baseline median | Candidate median | Improvement |
|---:|---:|---:|---:|
| 32 | 5.9163 ms | 2.8950 ms | 51.1% |
| 64 | 5.4144 ms | 2.5348 ms | 53.2% |
| 128 | 5.2771 ms | 3.3687 ms | 36.2% |
| 256 | 4.8534 ms | 2.3415 ms | 51.8% |
| 512 | 5.1568 ms | 3.0486 ms | 40.9% |
| 1,024 | 6.1759 ms | 2.9539 ms | 52.2% |
| 2,048 | 6.3326 ms | 2.4928 ms | 60.6% |
| 4,096 | 6.4489 ms | 2.4611 ms | 61.8% |
| 8,192 | 4.8515 ms | 2.4578 ms | 49.3% |
| 16,384 | 4.6563 ms | 3.3323 ms | 28.4% |
| 32,768 | 4.4568 ms | 2.4873 ms | 44.2% |
| 65,536 | 4.6797 ms | 3.6875 ms | 21.2% |
| 131,072 | 4.5995 ms | 3.4486 ms | 25.0% |

Median across the 13 per-size medians: **5.1568 ms → 2.8950 ms**. The median per-size speedup was **1.97x (49.3%)**. Criterion classified every case as improved with `p = 0.00`; per-case point improvements ranged from 21.2% to 61.8%.

## Correctness validation

The test build independently reconstructs

```text
[D]₁ + MSM(F terms) - E[G]₁ + z[W_z]₁ + uzω[W_zω]₁
```

and compares it with the regrouped right pairing input. Local mutation checks confirmed the regression fails for a wrong `E` sign, an omitted `z[W_z]₁` term, and swapped G2 pairing partners; all mutations were reverted.

Completed successfully:

- `make test` (release, default and all features)
- `make no-std`
- `make clippy`
- `make fmt`

### DIFF
--- a/src/proof_system/proof.rs
+++ b/src/proof_system/proof.rs
@@ -321,26 +321,6 @@ pub(crate) mod alloc {
                 &z_challenge,
             );
 
-            // Compute '[D]_1'
-            let D = self
-                .compute_linearization_commitment(
-                    &alpha,
-                    &beta,
-                    &gamma,
-                    (
-                        &range_sep_challenge,
-                        &logic_sep_challenge,
-                        &fixed_base_sep_challenge,
-                        &var_base_sep_challenge,
-                    ),
-                    &z_challenge,
-                    &u_challenge,
-                    l1_eval,
-                    verifier_key,
-                    &domain,
-                )
-                .0;
-
             // Evaluate public inputs
             let pi_eval = compute_barycentric_eval_sparse(
                 public_input_indexes,
@@ -417,8 +397,30 @@ pub(crate) mod alloc {
                 .sum();
             E_scalar += -r_0_eval + (u_challenge * self.evaluations.z_eval);
 
+            // Fold all terms of [F]_1 - [E]_1 and the opening-witness
+            // terms for the right pairing input into the linearization MSM.
+            let mut scalarmuls_points = Vec::with_capacity(32);
+            let mut scalarmuls_scalars = Vec::with_capacity(32);
+            self.append_linearization_commitment_terms(
+                &mut scalarmuls_scalars,
+                &mut scalarmuls_points,
+                &alpha,
+                &beta,
+                &gamma,
+                (
+                    &range_sep_challenge,
+                    &logic_sep_challenge,
+                    &fixed_base_sep_challenge,
+                    &var_base_sep_challenge,
+                ),
+                &z_challenge,
+                z_h_eval,
+                &u_challenge,
+                l1_eval,
+                verifier_key,
+            );
             let mut f_scalars = [BlsScalar::zero(); V_MAX_DEGREE];
-            f_scalars.clone_from_slice(&v_coeffs_E[..V_MAX_DEGREE]);
+            f_scalars.copy_from_slice(&v_coeffs_E[..V_MAX_DEGREE]);
 
             // As we include the shifted coefficients when computing [F]_1,
             // we group them to save scalar multiplications when multiplying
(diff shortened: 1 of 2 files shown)


src/proof_system/proof.rs
```diff
@@ -705,40 +737,68 @@ pub(crate) mod alloc {
...
+                ),
+                &f_points,
+                &f_scalars,
+                opening_key.g,
+                E_scalar,
+                self.w_z_chall_comm.0,
+                z_challenge,
+                self.w_z_chall_w_comm.0,
+                shifted_witness_scalar,
             );
```
moCello, line 792: **CAUTION:** The regrouped `verify_legacy` right pairing input is never executed by any test, so this equivalence assertion is dead code on the legacy path.

**Context:** `tests/plonk_versioning.rs` only asserts that V1 *proving* is unsupported; nothing calls `verify_with_version(.., PlonkVersion::V1)`. Codecov's 16 uncovered patch lines are this block. `cfg(test)` is only set when the library itself is compiled for unit tests, so an integration test under `tests/` would cover the legacy lines but still skip this assertion. The assertion is an algebraic identity that holds for any proof bytes, so it does not need a valid V1 proof to be meaningful — a V3 proof pushed through the V1 path reaches the pairing check and fails there.

**Fix:** In the inline `soundness_tests` module (the only place a proof is verified with `cfg(test)` on), add `verifier.verify_with_version(&honest_proof, &honest_pi, PlonkVersion::V1)` and assert `Err(Error::ProofVerificationError)`; that drives the legacy path end-to-end and makes the reference comparison actually run.

src/proof_system/proof.rs
```diff
@@ -443,40 +445,68 @@ pub(crate) mod alloc {
...
```
moCello, line 500: **NOTE:** This test-only assertion sits inside the production `verify` body, and no named test owns the equivalence it checks.

**Context:** Every proof verified from `tests/` and `benches/` runs the library without `cfg(test)` and never reaches it. The unit tests that do reach it — `soundness_tests::forged_selector_eval_proof_must_be_rejected` and the soundness suite's `assert_verifies` callers — reach it incidentally, as a side effect of verifying a proof. So a future edit deleting the block leaves no failing test behind, and the PR's mutation checks (wrong `E` sign, dropped `z[W_z]₁`) rest on that incidental coverage. The same shape in `verify_legacy` executes zero times.

**Fix:** Move the right-input assembly (`append_linearization_commitment_terms` + `f_points`/`f_scalars` + generator/witness pushes) into a private helper returning the scalar/point vectors, and assert equivalence against the reference from a named unit test in the existing `#[cfg(test)]` module, ideally over more than one circuit shape. Yours to take or leave; the legacy-path fix above is the minimum.

src/proof_system/proof.rs
```diff
@@ -818,32 +963,47 @@ pub(crate) mod alloc {
...
+        witness: G1Affine,
+        z_challenge: BlsScalar,
+        shifted_witness: G1Affine,
+        shifted_witness_scalar: BlsScalar,
+    ) {
+        let expected = reference_linearization
+            + msm_variable_base(f_points, f_scalars)
+            - generator * e_scalar
+            + witness * z_challenge
+            + shifted_witness * shifted_witness_scalar;
```
moCello, line 1000: **NOTE:** Two new `#[allow(clippy::too_many_arguments)]` attributes with no justifying comment — this one and the one on `compute_reference_linearization_commitment`; `make clippy` and CI clippy run `cargo clippy --features=rkyv/size_32` without `--tests`, so the lint never reaches either `cfg(test)` item and both attributes are inert.

**Fix:** Delete both attributes. Alternatively drop the helper and compute `expected` at the two call sites (`reference + msm_variable_base(&f_points, &f_scalars) − opening_key.g * E_scalar + w_z * z + w_zw * shifted`) followed by `assert_eq!`. Either rides the same push as the legacy-path test above.

moCello (commented):
COMMENT — The regrouped equation is algebraically identical to the old one (quotient scalars, `−E`, `z`, `u·z·ω` all trace back to the previous code) and the API, transcript and pairing order are untouched, but the legacy verifier was rewritten with no test executing it, and that gap should close in this PR since the fix is one verify call in the existing inline soundness test.
- One caution: add a `PlonkVersion::V1` verification call in the inline unit-test module (not `tests/`, where `cfg(test)` is off for the lib) so `verify_legacy` and its equivalence assertion actually run.
- Two notes: the in-body `cfg(test)` assertion shape and the new `too_many_arguments` allow.
- Process: commit `29ce0f8` is subject-only. `33357f5` and `1f67564` each carry a body stating what changed and why, closing with `Resolves #N`. Add one stating what moved into the single MSM and why the algebra is unchanged, plus `Resolves #920`.

src/proof_system/proof.rs
HDauven: Added the inline V1 verification call. The existing V3 proof now exercises verify_legacy with cfg(test) enabled and is required to return ProofVerificationError, so the legacy grouped/reference equivalence assertion runs end to end.

src/proof_system/proof.rs
HDauven: The V1 path is now exercised explicitly alongside the current verifier path. I kept the equivalence assertion at the assembly site rather than adding a larger production helper refactor solely for test ownership.

src/proof_system/proof.rs
HDauven: Removed both inert test-only too_many_arguments allowances.

moCello (approved):
APPROVE — every prior item that asked for a change is closed: the legacy path is now driven end-to-end by the V1 verification call, the inert lint allowances are gone, the commit carries its rationale, and the regrouped equation still reduces term-for-term to the previous separate right-input computation with transcript, pairing order and API untouched.
- The note on the equivalence assertion sitting in the production `verify` body stays open by your call; nothing else does.

PR #923, opened 2026-08-14
Optimize prover hot paths

### LINKED ISSUE #921: Improve prover hot-path performance
## Summary

The prover repeats avoidable work in several hot paths: sigma polynomials are transformed for every proof, permutation denominators are inverted individually, aggregate opening polynomials allocate intermediate values, and independent wire/quotient work is serialized in `std` builds.

## Acceptance criteria

- Reuse safely reconstructed sigma evaluations across proofs.
- Batch permutation denominator inversion without changing the product argument.
- Reduce aggregate-opening polynomial clones and temporary allocations.
- Parallelize only independent prover work behind the existing `std` feature.
- Preserve proof/transcript behavior, serialization compatibility, checked deserialization, and `no_std` support.
- Add equivalence/round-trip coverage and demonstrate an improvement with the repository benchmark.

HDauven:
## Summary

- cache sigma evaluations when a prover is constructed, rebuilding them from the checked prover key after deserialization
- replace per-row permutation inversions with one batch inversion while preserving the product argument
- build aggregate opening witnesses in one coefficient buffer and borrow their source polynomials
- parallelize independent wire polynomial/commitment and quotient commitment work in `std` builds while preserving RNG draw order
- parallelize independent butterfly chunks within large FFT stages in `std` builds; `no_std` keeps the serial implementation

The serialized prover format, proof mathematics, transcript, and RNG draw order are unchanged. No unchecked deserialization path is added. The changes are split into separate commits for the permutation, aggregate-witness, sigma-cache, independent prover-work, and FFT optimizations.

## Memory footprint

Each `Prover` now retains four sigma-evaluation vectors containing `4 * n` BLS scalars, or `128 * n` bytes. At a proving domain of `2^17`, this is 16 MiB per prover and per clone. A consumer retaining four such provers therefore adds approximately 64 MiB of steady-state memory. The cache is not serialized; checked deserialization reconstructs it from the validated prover key.

## Repository benchmark

The repository's Criterion benchmark was run on the exact base and candidate, sequentially on the same machine, for the 131,072-gate proving case. Criterion used its standard three-second warmup and ten measured iterations.

| Revision | Median proving time |
|---|---:|
| Base `768cf849` | 12.950 s |
| Candidate | 6.143 s |

The candidate reduced median proving time by **52.6%** in the repository-native benchmark.

## Phoenix 2^17 benchmark

The final FFT path was also benchmarked on the dedicated CPX42 reference worker (`dusk-compute-fsn1-11`, AMD EPYC Genoa, 8 shared vCPU, 16 GiB RAM) using the production-shaped four-input Phoenix transfer circuit:

- 131,024 constraints
- 131,072-element proving domain
- `RAYON_NUM_THREADS=8`
- one warmup and five measured proofs per run
- alternating order: previous head, amended head, amended head, previous head

| Revision | Run A median | Run B median |
|---|---:|---:|
| Previous PR head `4f5319ef` | 13.208 s | 13.199 s |
| Amended head `6aa1662b` | 8.830 s | 8.858 s |

The large-domain parallel FFT reduced Phoenix proving time by **32.9–33.2%** on the reference worker. Verification remained effectively unchanged, as expected for a prover-only change.

As a same-machine historical control, the exact April private optimization stack measured 3.642 s on an Intel Core Ultra 9 275HX. The current optimization stack with this FFT path measured 3.664 s for today's 2.1%-larger Phoenix circuit, indicating that the previously observed gap was caused by the omitted FFT path rather than the circuit-hardening work.

## Validation

- parallel FFT output equals the serial implementation on a large domain
- deterministic V3 proof bytes match the base revision
- `make test`
- `make no-std`
- `make clippy`
- `make fmt`
- `git diff --check`
- repository Criterion benchmark at 131,072 gates
- production-shaped Phoenix proof generation and verification on the CPX42 reference worker

Further late-stage FFT parallelism and helper deduplication are tracked in #933.

Resolves #921

### DIFF
--- a/src/compiler/prover.rs
+++ b/src/compiler/prover.rs
@@ -32,6 +32,7 @@ pub struct Prover {
     pub(crate) commit_key: CommitKey,
     pub(crate) verifier_key: VerifierKey,
     pub(crate) transcript: Transcript,
+    sigma_evaluations: [Vec<BlsScalar>; 4],
     pub(crate) size: usize,
     pub(crate) constraints: usize,
 }
@@ -52,19 +53,27 @@ impl Prover {
         verifier_key: VerifierKey,
         size: usize,
         constraints: usize,
-    ) -> Self {
+    ) -> Result<Self, Error> {
         let transcript =
             Transcript::base(label.as_slice(), &verifier_key, constraints);
+        let domain = EvaluationDomain::new(constraints)?;
+        let sigma_evaluations = [
+            domain.fft(&prover_key.permutation.s_sigma_1.0),
+            domain.fft(&prover_key.permutation.s_sigma_2.0),
+            domain.fft(&prover_key.permutation.s_sigma_3.0),
+            domain.fft(&prover_key.permutation.s_sigma_4.0),
+        ];
 
-        Self {
+        Ok(Self {
             label,
             prover_key,
             commit_key,
             verifier_key,
             transcript,
+            sigma_evaluations,
             size,
             constraints,
-        }
+        })
     }
 
     /// adds blinding scalars to a witness vector
@@ -222,14 +231,14 @@ impl Prover {
 
         let verifier_key = VerifierKey::from_slice(verifier_key)?;
 
-        Ok(Self::new(
+        Self::new(
             label,
             prover_key,
             commit_key,
             verifier_key,
             size,
             constraints,
-        ))
+        )
     }
 
     /// Prove the circuit using the current (latest) proving behavior.
@@ -370,10 +379,10 @@ impl Prover {
 
         let gamma = transcript.challenge_scalar(b"gamma");
         let sigma = [
-            &self.prover_key.permutation.s_sigma_1.0,
-            &self.prover_key.permutation.s_sigma_2.0,
-            &self.prover_key.permutation.s_sigma_3.0,
-            &self.prover_key.permutation.s_sigma_4.0,
+            self.sigma_evaluations[0].as_slice(),
+            self.sigma_evaluations[1].as_slice(),
+            self.sigma_evaluations[2].as_slice(),
+            self.sigma_evaluations[3].as_slice(),
         ];
         let wires = [
             a_scalars.as_slice(),
--- a/src/compiler/prover.rs
+++ b/src/compiler/prover.rs
@@ -595,21 +595,21 @@ impl Prover {
         // compute the opening proof polynomial 'W_z(X)'
         let aggregate_witness = CommitKey::compute_aggregate_witness(
             &[
-                r_poly,
-                a_poly.clone(),
-                b_poly.clone(),
-                c_poly,
-                d_poly.clone(),
-                self.prover_key.permutation.s_sigma_1.0.clone(),
-                self.prover_key.permutation.s_sigma_2.0.clone(),
-                self.prover_key.permutation.s_sigma_3.0.clone(),
+                &r_poly,
+                &a_poly,
+                &b_poly,
+                &c_poly,
+                &d_poly,
+                &self.prover_key.permutation.s_sigma_1.0,
+                &self.prover_key.permutation.s_sigma_2.0,
+                &self.prover_key.permutation.s_sigma_3.0,
                 // Bind selector evaluations (q_*) used inside the verifier
                 // linearization commitment to their committed polynomials
                 // by including them in the same batched opening at `z`.
-                self.prover_key.arithmetic.q_arith.0.clone(),
-                self.prover_key.arithmetic.q_c.0.clone(),
-                self.prover_key.arithmetic.q_l.0.clone(),
-                self.prover_key.arithmetic.q_r.0.clone(),
+                &self.prover_key.arithmetic.q_arith.0,
+                &self.prover_key.arithmetic.q_c.0,
+                &self.prover_key.arithmetic.q_l.0,
+                &self.prover_key.arithmetic.q_r.0,
             ],
             &z_challenge,
             &v_challenge,
@@ -621,7 +621,7 @@ impl Prover {
 
         // compute the shifted opening proof polynomial 'W_zw(X)'
         let shifted_aggregate_witness = CommitKey::compute_aggregate_witness(
-            &[z_poly, a_poly, b_poly, d_poly],
+            &[&z_poly, &a_poly, &b_poly, &d_poly],
             &(z_challenge * domain.group_gen),
             &v_w_challenge,
         );
--- a/src/compiler/prover.rs
+++ b/src/compiler/prover.rs
@@ -14,7 +14,7 @@ use merlin::Transcript;
 use rand_core::{CryptoRng, RngCore};
 
 use super::{Circuit, Composer, PlonkVersion};
-use crate::commitment_scheme::CommitKey;
+use crate::commitment_scheme::{CommitKey, Commitment};
 use crate::compiler::prover::linearization_poly::ProofEvaluations;
 use crate::error::Error;
 use crate::fft::{EvaluationDomain, Polynomial};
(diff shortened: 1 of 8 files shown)


src/compiler/prover.rs
```diff
@@ -55,13 +58,22 @@ impl Prover {
     ) -> Self {
         let transcript =
             Transcript::base(label.as_slice(), &verifier_key, constraints);
+        let domain = EvaluationDomain::new(constraints)
+            .expect("compiler constructed an invalid evaluation domain");
```
moCello: **CAUTION:** `Prover::new` panics on an invalid domain, and it now sits on the `try_from_bytes` deserialization path where the caller returns `Result`.

**Context:** `try_from_bytes` feeds `constraints` from untrusted bytes into `Prover::new`. Today the panic is unreachable only because `ProverKey::from_slice` has already built the `8n` domain (which fails first for any `n` where the `n` domain would fail) and `constraints.next_power_of_two() == size == prover_key.n` is checked. That ordering is held in another module with nothing near `Prover::new` naming it; a later change to `ProverKey::from_slice` or to the size checks re-arms the panic, and the message ("compiler constructed …") is wrong for the deserialization path. `Compiler::preprocess` already has the same domain in scope (`EvaluationDomain::new(size - 1)?`) and discards it before calling `Prover::new`.

**Severity-rationale:** Unreachable at this head, but the only guard is an ordering held in another module rather than anything the compiler enforces, and the call now sits on a path that takes untrusted bytes.

**Fix:** Make `Prover::new` return `Result<Self, Error>` and use `?` (both callers already return `Result`), or pass the already-constructed `EvaluationDomain` in from `Compiler::compile` and construct it with `?` in `try_from_bytes`.

src/compiler/prover.rs
```diff
@@ -345,16 +406,38 @@ impl Prover {
...
+        );
+
+        #[cfg(not(feature = "std"))]
         let a_poly = Self::blind_poly(rng, &a_scalars, 1, &domain);
+        #[cfg(not(feature = "std"))]
         let b_poly = Self::blind_poly(rng, &b_scalars, 1, &domain);
+        #[cfg(not(feature = "std"))]
         let c_poly = Self::blind_poly(rng, &c_scalars, 1, &domain);
+        #[cfg(not(feature = "std"))]
         let d_poly = Self::blind_poly(rng, &d_scalars, 1, &domain);
```
moCello: **NOTE:** Round 1 and the quotient commitment now have two separately-maintained call sites — 11 `#[cfg]` toggles in round 1 and 5 more at the quotient commitment — and the `no_std` one is compiled out of every test run.

**Context:** `make test` and CI run only with `std`; `make no-std` builds but never executes. The serial arm's RNG draw order and commitment order are therefore asserted by nothing — `precomputed_wire_blinders_preserve_rng_order` and `deterministic_v3_proof_matches_base_digest` only exercise the rayon arm, and the former does not compile without `std` since `sample_wire_blinders` is `std`-gated while the test module is not. Same shape at `commit_polynomials` (`prove_inner` lines 544-560). `sample_wire_blinders`, `blind_wire_polynomials` and `commit_polynomials` carry no `std`-specific signature; only their bodies do.

**Fix:** Keep one call site: drop the `#[cfg(feature = "std")]` on the three helpers and put the `cfg` inside them (rayon `join` under `std`, plain sequential calls otherwise). The `no_std` build then runs the same tested sequence with a different executor, and `precomputed_wire_blinders_preserve_rng_order` compiles under both feature sets.

src/composer/permutation.rs
```diff
@@ -304,6 +299,7 @@ impl Permutation {
 mod test {
     use dusk_bls12_381::BlsScalar;
     use ff::Field;
+    use itertools::izip;
```
moCello, line 302: **NOTE:** With the production `izip!` uses gone, `itertools` is used only inside `#[cfg(test)]` but is still a regular dependency and pulled into the `std` feature (`itertools/default`).

**Fix:** Move `itertools` to `[dev-dependencies]` and drop `"itertools/default"` from the `std` feature list; downstream consumers no longer compile it.

moCello (commented):
COMMENT — The optimizations are mathematically equivalent to the previous code (product argument, aggregate witness, RNG order and proof bytes all pinned by tests) and nothing gates the merge. One caution asks for a change on this branch; the four notes ride the same push.
- That caution is the `expect` in `Prover::new`, which now sits on the `try_from_bytes` path taking untrusted bytes; returning `Result` removes the class.
- The `sigma_evaluations` cache adds `4·n·32` bytes of resident memory per `Prover` (16 MiB at 2^17) and to every `Clone`; `rusk-prover` holds four Phoenix provers via `try_from_bytes`, so that is tens of MiB of new steady-state footprint there — worth a line in the PR description.
- Process, non-blocking: the single commit bundles five separable optimizations (sigma cache, batch inversion, aggregate witness, rayon joins, parallel FFT); the FFT change in particular would bisect more cleanly as its own commit. The linked issue asks for an improvement shown with the repository benchmark; the PR reports an out-of-repo Phoenix measurement instead — a `cargo bench` before/after line would close that gap.

src/fft/domain.rs
```diff
@@ -355,10 +355,51 @@ pub(crate) mod alloc {
...
+            let chunk_count = n / chunk_len;
+
+            if chunk_count >= PARALLEL_FFT_MIN_CHUNKS {
+                a.par_chunks_mut(chunk_len)
+                    .for_each(|chunk| butterfly_chunk(chunk, m, w_m));
+            } else {
+                for chunk in a.chunks_mut(chunk_len) {
+                    butterfly_chunk(chunk, m, w_m);
+                }
+            }
```
moCello, line 397: **NOTE:** The last two butterfly stages (`chunk_count` 1 and 2) and the bit-reversal pass stay serial, which caps the FFT speedup at roughly half the thread count on the 2^17–2^20 domains this PR targets; a tracked follow-up, not for this PR.

**Context:** Each stage does `n/2` butterflies, so the two serial stages are `2/log_n` of the butterfly work (~10–12% at 2^17–2^20) plus the O(n) serial bit-reverse. By Amdahl that leaves ~0.23 of serial time at 8 threads against an ideal 0.125. Inside a chunk the `j` loop is itself independent per `j`; a sub-range starting at `j0` only needs `w = w_m.pow(j0)`. The bit-reverse loop and the butterfly body are also verbatim copies of the ones in `serial_fft`, so the same algorithm now lives in two places.

**Fix:** In a follow-up: when `chunk_count < PARALLEL_FFT_MIN_CHUNKS`, split `left`/`right` of each chunk into equal sub-ranges via `par_chunks_mut` and seed each sub-range's twiddle with `w_m.pow(offset)`; hoist the bit-reversal loop into a shared helper and have `serial_fft` iterate `chunks_mut(2 * m)` over `butterfly_chunk` so both paths share one butterfly.

src/compiler/prover.rs
```diff
@@ -898,6 +994,92 @@ mod tests {
...
+            .expect("serialized prover should decode");
+        let domain = EvaluationDomain::new(decoded.constraints).unwrap();
+        let expected = [
+            domain.fft(&decoded.prover_key.permutation.s_sigma_1.0),
+            domain.fft(&decoded.prover_key.permutation.s_sigma_2.0),
+            domain.fft(&decoded.prover_key.permutation.s_sigma_3.0),
+            domain.fft(&decoded.prover_key.permutation.s_sigma_4.0),
+        ];
+
+        assert_eq!(decoded.sigma_evaluations, expected);
```
moCello: **NOTE:** The roundtrip assertion recomputes the constructor's own expression from the decoded key, so it passes for any `Prover::new` that FFTs whatever key it is given; it does not check the cache matches the compiled prover.

**Fix:** Assert `decoded.sigma_evaluations == prover.sigma_evaluations` (the compile-time cache is in scope as `prover`) — that pins the cache across the serialize/deserialize boundary rather than against itself.

src/compiler/prover.rs
HDauven: Prover::new now returns Result and propagates EvaluationDomain::new failures. Both compilation and checked prover deserialization use the fallible constructor.

src/compiler/prover.rs
HDauven: Consolidated both duplicated call sites. The blinder, wire-polynomial, and commitment helpers are available in both configurations, with only their rayon/sequential executor bodies gated; std and no_std now share RNG and commitment ordering.

src/composer/permutation.rs
HDauven: Moved itertools to dev-dependencies and removed it from the std feature. The remaining uses are confined to permutation tests.

src/fft/domain.rs
HDauven: Tracked the late-stage parallelism and shared-helper work separately in #933, including twiddle-offset tests and repository benchmark acceptance criteria.

src/compiler/prover.rs
HDauven: Changed the regression to compare the decoded sigma cache directly with the compiled prover cache.

xevisalle (approved):

moCello (approved):
APPROVE — every point from the previous round is addressed on this head: fallible `Prover::new` on the deserialization path, one shared call sequence for `std` and `no_std` with only executor bodies gated, `itertools` demoted to a dev-dependency, the roundtrip test pinning the decoded sigma cache against the compiled one, and the FFT tail tracked in #933; the sigma cache is built on the same domain the prover uses and its size is bounded by the validated key, the batched permutation product and single-buffer aggregate witness are algebraically identical to the previous code, the parallel FFT reproduces the serial butterfly order and is pinned by an equivalence test that hits both its parallel and serial-tail arms, and the base-revision proof digest test locks proof bytes and RNG order.

PR #924, opened 2026-08-14
Cache deterministic quotient-domain prover data

### LINKED ISSUE #919: Cache deterministic quotient-domain prover data
Quotient construction rebuilds deterministic domain data on every proof, including the quotient evaluation domain, inverses of the fixed vanishing-polynomial coset evaluations, and first-Lagrange coset evaluations. These values can instead be derived once when a `Prover` is constructed.

Acceptance criteria:

- Cache only deterministic quotient-domain data derived from the circuit size and validated prover key.
- Rebuild derived data after checked prover deserialization rather than changing the serialized format.
- Preserve proof bytes, transcript challenges, verifier-key compatibility, and public API behavior.
- Keep malformed prover-key data rejected and add no unchecked deserialization API.
- Preserve feature and `no_std`/`alloc` builds.
- Benchmark proving against the current `master` and document both the speedup and memory cost.

HDauven:
## Summary

- cache the base and `8n` quotient evaluation domains on `Prover`
- cache the eight-value period of the validated vanishing-coset inverses
- derive first-Lagrange coset evaluations from their closed form with one batch inversion per proof instead of retaining another `8n` vector
- rebuild all retained derived state after checked deserialization without changing prover serialization

## Safety and compatibility

The caches are private derived state and are not serialized. `Prover::try_from_bytes` still performs checked prover-key decoding, including exact evaluation-domain, linear-coset and vanishing-evaluation validation, before cache construction. The constructor also rejects inconsistent circuit sizes, wrong cache-source lengths and zero vanishing evaluations before they can become proving state.

The first-Lagrange values are reconstructed as `L1(x) = (x^n - 1) / (n * (x - 1))` from those validated coset evaluations. This adds no unchecked API and does not change proof encoding, transcript inputs, verifier keys or serialized prover bytes. The deterministic V3 proof digest remains byte-identical to the public base behavior.

## Performance

Measured against current `master` at `0630625c06f916f79804be765b8eb2a47c4bb6a3` on the designated CPX42 reference worker. The workload is a production-shaped four-input Phoenix transfer with 131,024 constraints and a `2^17` proving domain. Each variant ran in three separate processes, with one warm-up and one measured proof per process; order was rotated across rounds.

| Variant | Median proof | Proof range | Median peak RSS |
| --- | ---: | ---: | ---: |
| current `master` | 52.141s | 51.737–53.058s | 2,151,112 KiB |
| retain full `8n` L1 vector | 40.204s | 40.124–40.953s | 2,184,020 KiB |
| this PR: closed-form L1 | 40.660s | 40.529–41.025s | 2,151,372 KiB |

The selected design reduces median proving time by **22.02%** relative to current `master`. Retaining the full vector was only 1.13% faster, but added 32.14 MiB of measured peak RSS. The selected closed form was within 0.25 MiB of baseline peak RSS and avoids permanently retaining a 32 MiB vector per `2^17` prover.

Across Rusk's one `2^16` and three `2^17` Phoenix provers, retaining that vector would cost exactly 112 MiB. Compilation and verification showed no material change between variants.

## Validation

- rebased directly onto current `master`, preserving the merged #923 and #926 implementations
- `make test`
- `make fmt`
- `make clippy`
- `make no-std`
- focused checked-deserialization, malformed-constructor and cache round-trip tests
- deterministic V3 proof digest compatibility fixture
- three-way production-shaped Phoenix benchmark on the CPX42 reference worker

Resolves #919

### DIFF
--- a/src/compiler/prover.rs
+++ b/src/compiler/prover.rs
@@ -23,6 +23,7 @@ use crate::proof_system::{
     ProverKey, VerifierKey, linearization_poly, quotient_poly,
 };
 use crate::transcript::TranscriptProtocol;
+use crate::util::batch_inversion;
 
 /// Turbo Prover with processed keys
 #[derive(Clone)]
@@ -33,6 +34,9 @@ pub struct Prover {
     pub(crate) verifier_key: VerifierKey,
     pub(crate) transcript: Transcript,
     sigma_evaluations: [Vec<BlsScalar>; 4],
+    domain: EvaluationDomain,
+    quotient_domain: EvaluationDomain,
+    vanishing_coset_inverses: [BlsScalar; 8],
     pub(crate) size: usize,
     pub(crate) constraints: usize,
 }
@@ -54,9 +58,40 @@ impl Prover {
         size: usize,
         constraints: usize,
     ) -> Result<Self, Error> {
+        if constraints.checked_next_power_of_two() != Some(size)
+            || prover_key.n != size
+        {
+            return Err(dusk_bytes::Error::InvalidData.into());
+        }
+
+        let domain = EvaluationDomain::new(constraints)?;
+        let quotient_size = domain
+            .size()
+            .checked_mul(8)
+            .ok_or(dusk_bytes::Error::InvalidData)?;
+        let quotient_domain = EvaluationDomain::new(quotient_size)?;
+
+        // Compilation constructs these evaluations for this exact quotient
+        // domain, while checked decoding validates their domain, length, and
+        // values before construction reaches here. Keep the checks local too,
+        // so malformed source data can never become cached proving state.
+        let vanishing_evaluations = &prover_key.v_h_coset_8n().evals;
+        if vanishing_evaluations.len() != quotient_domain.size()
+            || vanishing_evaluations
+                .iter()
+                .any(|evaluation| *evaluation == BlsScalar::zero())
+        {
+            return Err(dusk_bytes::Error::InvalidData.into());
+        }
+        let first_period = vanishing_evaluations
+            .get(..8)
+            .ok_or(dusk_bytes::Error::InvalidData)?;
+        let mut vanishing_coset_inverses = [BlsScalar::zero(); 8];
+        vanishing_coset_inverses.copy_from_slice(first_period);
+        batch_inversion(&mut vanishing_coset_inverses);
+
         let transcript =
             Transcript::base(label.as_slice(), &verifier_key, constraints);
-        let domain = EvaluationDomain::new(constraints)?;
         let sigma_evaluations = [
             domain.fft(&prover_key.permutation.s_sigma_1.0),
             domain.fft(&prover_key.permutation.s_sigma_2.0),
@@ -71,6 +106,9 @@ impl Prover {
             verifier_key,
             transcript,
             sigma_evaluations,
+            domain,
+            quotient_domain,
+            vanishing_coset_inverses,
             size,
             constraints,
         })
@@ -174,6 +212,8 @@ impl Prover {
     fn prepare_serialize(
         &self,
     ) -> (usize, Vec<u8>, Vec<u8>, [u8; VerifierKey::SIZE]) {
+        // Quotient caches are derived state. Omitting them preserves the
+        // format and ensures checked decoding always rebuilds fresh values.
         let prover_key = self.prover_key.to_var_bytes();
         let commit_key = self.commit_key.to_raw_var_bytes();
         let verifier_key = self.verifier_key.to_bytes();
@@ -384,10 +424,8 @@ impl Prover {
     {
         let prover = Composer::prove(self.constraints, circuit)?;
 
-        let constraints = self.constraints;
         let size = self.size;
-
-        let domain = EvaluationDomain::new(constraints)?;
+        let domain = self.domain;
 
         let mut transcript = self.transcript_for_version(version);
 
@@ -494,11 +532,12 @@ impl Prover {
             var_base_sep_challenge,
         );
         let t_poly = quotient_poly::compute(
-            &domain,
+            &self.quotient_domain,
             &self.prover_key,
             &z_poly,
             wires,
             &pi_poly,
+            &self.vanishing_coset_inverses,
             args,
         )?;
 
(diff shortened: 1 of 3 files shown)


moCello (commented):
COMMENT — The change is correct and format-preserving: proof bytes and serialized provers are unchanged, malformed keys still reject, and the public API is untouched. Nothing gates the merge; one caution asks for a change on this branch before the memory trade-off is locked in.
- That caution halves the memory the caches add, at no cost to the speedup. Three notes below — a redundant `domain` parameter on `compute`, untested error branches in `Prover::new`, a qualified path in the test — ride the same push if it is taken.
- Commit `943745f` is subject-only. `33357f5` and `1f67564` each carry a body stating what changed and why, closing with `Resolves #N` — worth matching on the next push.

src/compiler/prover.rs
```diff
@@ -52,19 +57,59 @@ impl Prover {
...
+        let mut vanishing_coset_inverses =
+            prover_key.v_h_coset_8n().evals.clone();
+        if vanishing_coset_inverses.len() != quotient_domain.size()
+            || vanishing_coset_inverses
+                .iter()
+                .any(|evaluation| *evaluation == BlsScalar::zero())
+        {
+            return Err(dusk_bytes::Error::InvalidData.into());
+        }
+        batch_inversion(&mut vanishing_coset_inverses);
```
moCello: **CAUTION:** The vanishing-coset inverse cache holds 8n scalars but only 8 distinct values; half of the added payload — 112 MiB of the 224 MiB the PR measures across the four Rusk provers — buys nothing.

**Context:** `v_h_coset_8n[i] = g^n · (ω_{8n}^n)^i − 1` and `ω_{8n}^n` is a primitive 8th root of unity, so the sequence repeats with period 8 (`EvaluationDomain::vanishing_poly_over_coset`: `step = group_gen^n`). Both construction paths guarantee that structure: `compile` computes the evaluations and `ProverKey::from_slice` checks them exactly. The per-element `invert()` this PR removes was the dominant cost; the same speedup comes from 8 inversions per prover indexed as `vanishing_coset_inverses[i & 7]` in `quotient_poly::compute`. That the `ProverKey` itself serializes 8n copies is pre-existing and out of scope.

**Impact:** The PR measures 234 MiB retained on `rusk-prover`; an 8-element cache cuts that to the Lagrange half with identical proof output.

**Fix:** Store `[BlsScalar; 8]` (take the first 8 evaluations after the existing length and nonzero checks), invert those, and index by `i & 7` in the quotient loop. Three places follow: the `debug_assert_eq!` on `vanishing_coset_inverses.len()` in `quotient_poly::compute`, and the length assertion and per-element zip in `prover_rebuilds_quotient_cache_after_checked_round_trip`. Update the memory numbers in the PR description accordingly.

src/proof_system/quotient_poly.rs
```diff
@@ -18,6 +18,7 @@ use crate::proof_system::ProverKey;
 /// [`ProverKey`] and some other info.
 pub(crate) fn compute(
     domain: &EvaluationDomain,
+    quotient_domain: &EvaluationDomain,
```
moCello: **NOTE:** `compute` now takes both domains, but `domain` is read only for `7 * domain.size()` and for a `debug_assert_eq!` that the two agree — an invariant the release build drops.

**Context:** the parameter list goes from six to nine. After the change `domain`'s only surviving use is the unsatisfied-circuit degree floor near the end of the function; every other reference became `quotient_domain`. `quotient_domain.size() / 8` gives the same value, so both the parameter and the cross-argument invariant it introduces can go. The invariant is soundness-adjacent: if the two ever disagreed, the degree bound would be wrong, and the only thing checking them is a `debug_assert_eq!`.

**Fix:** drop the `domain` parameter and derive the base size as `quotient_domain.size() / 8`; the first `debug_assert_eq!` then has nothing left to check. If bundling the cached values into one struct is preferred instead, define it in `proof_system` rather than passing `&Prover`, or the layering inverts.

src/compiler/prover.rs
```diff
@@ -674,6 +722,93 @@ mod tests {
...
+        let cached_scaled: Vec<_> = decoded
+            .first_lagrange_coset_evaluations
+            .iter()
+            .map(|evaluation| *evaluation * alpha.square())
+            .collect();
+        assert_eq!(cached_scaled, expected);
+
+        let mut malformed_key = decoded.prover_key.clone();
+        malformed_key.v_h_coset_8n.evals.pop();
+        let malformed = Prover::new(
```
moCello: **NOTE:** Of the error branches added to `Prover::new`, only the length-mismatch branch has a test; the size/`n` mismatch and zero-evaluation branches are uncovered.

**Context:** Both `ProverKey::from_slice` and `Compiler::compile` make those branches unreachable through the public API, so they are testable only via direct `Prover::new` calls like this one. The zero-evaluation branch matters most: `batch_inversion` silently skips zeros, so a regression that dropped the check would produce a valid-looking prover with a wrong quotient.

**Fix:** Add two more direct `Prover::new` cases in this module: one with `v_h_coset_8n.evals[0] = BlsScalar::zero()` and one with mismatched `size`/`constraints`, each asserting `Error::BytesError(InvalidData)`. Consider moving the malformed-key assertions out of `prover_rebuilds_quotient_cache_after_checked_round_trip` into their own test so each failure names one cause.

src/compiler/prover.rs
```diff
@@ -674,6 +722,93 @@ mod tests {
...
+
+        let alpha = BlsScalar::from(13u64);
+        let mut scaled_lagrange_evaluations =
+            vec![BlsScalar::zero(); decoded.domain.size()];
+        scaled_lagrange_evaluations[0] = alpha.square();
+        decoded
+            .domain
+            .ifft_in_place(&mut scaled_lagrange_evaluations);
+        let expected = decoded.quotient_domain.coset_fft(
+            &crate::fft::Polynomial::from_coefficients_vec(
```
moCello: **NOTE:** Fully qualified path in a function body; the test module already has a `use crate::fft::EvaluationDomain;` line to extend.

**Fix:** Import `Polynomial` alongside `EvaluationDomain` in the test module and call `Polynomial::from_coefficients_vec(...)`.

src/compiler/prover.rs
HDauven: Addressed in c5126bff: the vanishing inverse cache is now `[BlsScalar; 8]`, and quotient evaluation indexes the period with `i & 7`. The description now reports the exact 112 MiB + 1 KiB aggregate payload and marks the old RSS/load figures as superseded.

src/proof_system/quotient_poly.rs
HDauven: Addressed in c5126bff: `quotient_poly::compute` no longer accepts the base domain and derives the degree threshold from `quotient_domain.size() / 8`.

src/compiler/prover.rs
HDauven: Added in c5126bff: malformed constructor inputs now have a dedicated regression covering short evaluations, a zero evaluation, and inconsistent size/constraints.

src/compiler/prover.rs
HDauven: Cleaned up in c5126bff by importing `Polynomial` alongside `EvaluationDomain`.

xevisalle (approved):

src/compiler/prover.rs
```diff
@@ -32,6 +33,10 @@ pub struct Prover {
     pub(crate) commit_key: CommitKey,
     pub(crate) verifier_key: VerifierKey,
     pub(crate) transcript: Transcript,
+    domain: EvaluationDomain,
+    quotient_domain: EvaluationDomain,
+    vanishing_coset_inverses: [BlsScalar; 8],
+    first_lagrange_coset_evaluations: Vec<BlsScalar>,
```
moCello: **NOTE:** This `8n`-scalar cache has a closed form over data the prover key already holds, so it can be derived per proof rather than stored — 112 MiB across the four Phoenix keys.

**Context:** `L₁(g·ω₈ₙⁱ) = v_h_coset_8n[i & 7] · n⁻¹ · (linear_evaluations[i] − 1)⁻¹`, and both inputs already sit on `ProverKey`. Deriving it costs one `8n` `batch_inversion` per proof and drops the vector, which is roughly 4% on top of the ~190n-scalar prover key. The other three values cached here are `Copy` domains and eight scalars, so this is the only one with a material footprint.

**Fix:** Derive it inside `compute_permutation_checks` from `v_h_coset_8n` and `permutation.linear_evaluations` instead of storing it, if the memory outweighs one `batch_inversion` per proof.

moCello (approved):
APPROVE — Every point from the earlier round is addressed in `c5126bf`: the vanishing inverse cache is the eight-value period indexed by `i & 7`, `quotient_poly::compute` derives the degree floor from the quotient domain alone, the constructor's error branches each have a direct test, and the history is one commit with a body. Proof bytes, serialized provers, transcript inputs, and the public API are unchanged; nothing gates and nothing has to change on this branch.

- One note below, on the memory the new first-Lagrange cache costs against the closed form it could be derived from. Yours to weigh — it does not hold up the merge.

PR #926, opened 2026-08-15
Parallelize quotient coset FFTs

### LINKED ISSUE #925: Parallelize independent quotient coset FFTs under std
## Context

Quotient construction computes the coset FFTs for `z`, `a`, `b`, `c`, and `d` independently but currently runs them one after another. Native proving builds already enable Plonk's `std` feature through Rusk, while portable Phoenix/browser builds intentionally remain `no_std`.

## Proposed change

- Evaluate the five independent quotient-domain polynomials concurrently only when `std` is enabled.
- Keep the existing serial implementation for `no_std`.
- Preserve input/output ordering, transcript order, proof format, public APIs, and the underlying arithmetic.

## Evidence

On a production-shaped four-input Phoenix transfer circuit (131,024 constraints, 2^17 domain) on an 8-vCPU CPX42, alternating A/B runs measured:

- public `master`: 15.778-15.794 seconds median per proof
- candidate on public `master`: 12.779-12.809 seconds median per proof
- reduction: 18.953%

When combined with #923 and #924, the same candidate reduced the stacked result from 10.973-11.002 seconds to 8.027-8.097 seconds, or 26.628%. Verification and peak RSS were unchanged in both comparisons.

## Acceptance criteria

- Exact ordered equivalence with the serial coset FFT results.
- Full default and all-feature tests pass.
- Both embedded `no_std` build configurations pass without pulling Rayon into the portable path.
- No public API, transcript, proof-format, or serialization change.

HDauven:
Resolves #925

## Summary

- compute the five independent quotient coset FFTs concurrently when `std` is enabled
- retain the existing ordered serial path for `no_std` builds
- pin exact output equivalence and ordering with a focused regression

This does not change the quotient arithmetic, transcript, proof format, serialization, or public API. Native Rusk proving enables `dusk-plonk/std`; portable Phoenix and browser builds remain `no_std` and do not pull in Rayon.

## Phoenix benchmark

Production-shaped four-input transfer, 131,024 constraints, 2^17 proving domain, CPX42 with eight shared vCPUs, `RAYON_NUM_THREADS=8`. Each pass used one warm-up and five measured proofs; the order was baseline, candidate, candidate, baseline.

Directly against public `master`:

| Variant | Median pass A | Median pass B | Average median |
| --- | ---: | ---: | ---: |
| `master` | 15.794 s | 15.778 s | 15.786 s |
| Candidate | 12.809 s | 12.779 s | 12.794 s |

That is an 18.953% reduction. Verification latency and peak RSS were unchanged.

On top of #923 and #924, the same change reduced the average median from 10.988 s to 8.062 s, a 26.628% reduction.

## Validation

- `make test`
- `make clippy`
- `make no-std`
- `make doc`
- nightly formatting and diff checks
- mutation check: swapping two parallel outputs makes the ordering regression fail
- every measured Phoenix proof was verified by the unchanged verifier

### DIFF
--- a/src/proof_system/quotient_poly.rs
+++ b/src/proof_system/quotient_poly.rs
@@ -48,12 +48,16 @@ pub(crate) fn compute(
     // Compute 8n evals
     let domain_8n = EvaluationDomain::new(8 * domain.size())?;
 
-    let mut z_eval_8n = domain_8n.coset_fft(z_poly);
-
-    let mut a_eval_8n = domain_8n.coset_fft(a_poly);
-    let mut b_eval_8n = domain_8n.coset_fft(b_poly);
-    let c_eval_8n = domain_8n.coset_fft(c_poly);
-    let mut d_eval_8n = domain_8n.coset_fft(d_poly);
+    let [
+        mut z_eval_8n,
+        mut a_eval_8n,
+        mut b_eval_8n,
+        c_eval_8n,
+        mut d_eval_8n,
+    ] = compute_coset_evaluations(
+        &domain_8n,
+        [z_poly, a_poly, b_poly, c_poly, d_poly],
+    );
 
     for i in 0..8 {
         z_eval_8n.push(z_eval_8n[i]);
@@ -132,6 +136,26 @@ pub(crate) fn compute(
     }
 
     Ok(quotient_poly)
+}
+
+fn compute_coset_evaluations(
+    domain: &EvaluationDomain,
+    polynomials: [&Polynomial; 5],
+) -> [Vec<BlsScalar>; 5] {
+    #[cfg(not(feature = "std"))]
+    let evaluations = polynomials.map(|poly| domain.coset_fft(poly));
+
+    #[cfg(feature = "std")]
+    let evaluations = {
+        let mut evaluations: [Vec<BlsScalar>; 5] = Default::default();
+        evaluations
+            .par_iter_mut()
+            .zip(&polynomials[..])
+            .for_each(|(slot, poly)| *slot = domain.coset_fft(poly));
+        evaluations
+    };
+
+    evaluations
 }
 
 // Ensures that the circuit is satisfied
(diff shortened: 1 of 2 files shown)


src/proof_system/quotient_poly.rs
```diff
@@ -134,6 +138,24 @@ pub(crate) fn compute(
...
+    #[cfg(not(feature = "std"))]
+    let evaluations = polynomials.map(|poly| domain.coset_fft(poly));
+
+    #[cfg(feature = "std")]
+    let evaluations = polynomials
+        .into_par_iter()
+        .map(|poly| domain.coset_fft(poly))
+        .collect::<Vec<_>>()
+        .try_into()
+        .unwrap_or_else(|_| unreachable!("five inputs produce five outputs"));
```
moCello: **CAUTION:** Introduces an `unreachable!` panic path into library code, to discharge a `Result` that exists only because the array is round-tripped through `Vec`.

**Context:** The length is guaranteed by rayon's indexed `collect`, so the branch is dead, and it stays dead: the call site destructures five bindings from a five-argument call, and the new ordering test infers its array length from this signature, so changing one count alone fails to compile in two places. What is left is an unchecked-failure call in library code standing in for a conversion that does not need one.

**Fix:** Write directly into a fixed-size output so no length check is needed:

```rust
#[cfg(feature = "std")]
let evaluations = {
    let mut evaluations: [Vec<BlsScalar>; 5] = Default::default();
    evaluations
        .par_iter_mut()
        .zip(&polynomials[..])
        .for_each(|(slot, poly)| *slot = domain.coset_fft(poly));
    evaluations
};
```

`zip` on two indexed parallel iterators keeps position `i` paired with input `i`, so the existing ordering test still pins the behaviour. Zipping the slice rather than the array matters: array `IntoParallelIterator` arrived in rayon 1.5.1, and `.zip(polynomials)` would not build against the `rayon = "1.3"` the manifest declares.

moCello (commented):
COMMENT — The parallelisation is sound: the five transforms are independent, the coset FFT path is serial so nothing nests, and `no_std` stays rayon-free. Ordering is preserved by rayon's indexed collect and pinned by the new test, and the CHANGELOG entry is in place. Nothing gates the merge; one caution below asks for a change on this branch.
- Commit `fa03f16` is subject-only. `33357f5` and `1f67564` each carry a body stating what changed and why, closing with `Resolves #N` — worth matching on the next push.

src/proof_system/quotient_poly.rs
HDauven: Addressed in d93ff987: the parallel path now writes directly into a fixed `[Vec<BlsScalar>; 5]` through indexed `par_iter_mut().zip(&polynomials[..])`, removing both the fallible conversion and the panic path.

xevisalle (approved):

moCello (approved):
APPROVE — the fixed-size output removes the fallible conversion and the panic path flagged last round, and nothing new stands against the change.

`par_iter_mut().zip(&polynomials[..])` pairs indexed position `i` with input `i`, so the ordering test pins what it claims, and zipping the slice keeps the code building against the `rayon = "1.3"` floor the manifest declares. `coset_fft` is pure over a `Copy` domain of field constants, so the five tasks cannot observe each other and proof bytes stay independent of thread count. `best_fft` dispatches to `serial_fft`, so these five transforms are the only parallelism in this stage and nothing nests. Both arms of the helper are compile-guarded against an arity change, one by `array::map` and one by the explicit array annotation. Quotient arithmetic, transcript order, proof format and public API are unchanged. The commit now carries a body and `Resolves #925`.

Nothing gates the merge and nothing asks for a change on this branch.

PR #928, opened 2026-08-15
Cache verifier public-input roots

### LINKED ISSUE #927: Cache verifier domains and public-input roots
## Problem

Each proof verification rebuilds the fixed evaluation domain, recomputes every public-input root, and performs the first-Lagrange and sparse public-input inversions separately. These values and roots are deterministic for a compiled verifier.

## Scope

- Cache the evaluation domain and public-input roots when the verifier is compiled or checked-deserialized.
- Omit derived caches from serialization so the byte format stays unchanged.
- Reuse the existing vanishing-polynomial evaluation.
- Batch the first-Lagrange denominator and nonzero public-input denominators in one inversion pass.
- Preserve the current and legacy proof equations, transcripts, and public API.
- Return a verification error for a singular evaluation point instead of reaching an unwrap.

## Acceptance criteria

- Differential coverage against the existing Lagrange and sparse public-input formulas.
- Checked verifier round trips rebuild valid caches and preserve serialized bytes.
- Default, all-feature, Clippy, and no-std validation passes.
- A reference benchmark demonstrates a repeatable verifier improvement without a material proving or memory regression.

## Reference result

On the designated CPX42 worker, a production-shaped four-input Phoenix proof at a `2^17` domain improved from an interleaved average median of 5.5491 ms to 4.9137 ms across 2,000 verifications per pass, an 11.45% reduction. Compilation, proving, and peak RSS were unchanged within run variance.

HDauven:
## Summary

- cache the fixed evaluation domain and public-input roots in each compiled or checked-deserialized verifier
- reuse the vanishing-polynomial evaluation and batch the first-Lagrange and nonzero public-input denominators into one inversion pass
- preserve verifier serialization, proof/transcript equations, and the public API
- return `ProofVerificationError` for singular evaluation points instead of reaching the previous inversion unwrap
- reject evaluation-domain sizes that cannot be represented without unwinding during checked deserialization

## Performance

On the designated CPX42 reference worker, a production-shaped four-input Phoenix transfer proof at a `2^17` domain was measured in interleaved base/candidate/candidate/base order with 2,000 verifications per pass:

| Variant | Average median verification |
| --- | ---: |
| Grouped-MSM verifier | 5.5491 ms |
| This PR | 4.9137 ms |

That is an 11.45% reduction. The two individual comparisons were 11.05% and 11.85%. Compilation, proving, peak RSS, proof bytes, and verifier serialization were unchanged within run variance.

Two bounded MSM follow-ups were rejected rather than included: a larger 30-point MSM window was 19.5% slower in isolation, and splitting fixed points into cached WNAF tables was roughly 3.5x slower than the grouped MSM.

## Validation

- rebased directly onto current `master`
- full default and all-feature release tests
- Clippy and `no_std` builds
- formatting and diff checks
- differential coverage against the old separate L1 and sparse public-input formulas
- singular-point regressions for both `z = 1` and a nonzero public-input domain point
- debug regression proving oversized checked verifier input returns an error without unwinding
- mutation check: removing the domain-size inverse fails the differential regression
- real four-input Phoenix proofs generated and verified on both benchmark variants

Resolves #927

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remove the `Copy` derive from `Error` [#884]
 - Speed up proof verification by grouping the final pairing-input MSM terms
   and batch-normalizing its G1 inputs [#920]
+- Cache verifier domains and public-input roots and batch their denominator
+  inversions during proof verification [#927]
 - Change `component_add_point`, `component_sub_point` and `component_mul_point` to `TorsionFreeWitnessPoint` [#870]
 - Change `component_neg_point` and `component_select_identity` to `TorsionFreeWitnessPoint` [#870]
 - Change `component_select_identity` to constrain its `bit` to boolean [#870]
@@ -79,6 +81,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Reject unrepresentable evaluation domain sizes during verifier construction
+  and checked deserialization [#927]
 - Reject invalid witness, polynomial, selector and public-input indices when
   deserializing compressed circuits [#930]
 - Reject trailing data when deserializing compressed circuits [#913]
@@ -768,6 +772,7 @@ is necessary since `rkyv/validation` was required as a bound.
 [#930]: https://github.com/dusk-network/plonk/issues/930
 [#925]: https://github.com/dusk-network/plonk/issues/925
 [#921]: https://github.com/dusk-network/plonk/issues/921
+[#927]: https://github.com/dusk-network/plonk/issues/927
 [#920]: https://github.com/dusk-network/plonk/issues/920
 [#919]: https://github.com/dusk-network/plonk/issues/919
 [#916]: https://github.com/dusk-network/plonk/issues/916
--- a/src/compiler/verifier.rs
+++ b/src/compiler/verifier.rs
@@ -13,6 +13,7 @@ use merlin::Transcript;
 use super::PlonkVersion;
 use crate::commitment_scheme::OpeningKey;
 use crate::error::Error;
+use crate::fft::EvaluationDomain;
 use crate::proof_system::{Proof, VerifierKey};
 use crate::transcript::TranscriptProtocol;
 /// Verify proofs of a given circuit
@@ -21,6 +22,8 @@ pub struct Verifier {
     verifier_key: VerifierKey,
     opening_key: OpeningKey,
     public_input_indexes: Vec<usize>,
+    public_input_roots: Vec<BlsScalar>,
+    domain: EvaluationDomain,
     transcript: Transcript,
     size: usize,
     constraints: usize,
@@ -34,19 +37,26 @@ impl Verifier {
         public_input_indexes: Vec<usize>,
         size: usize,
         constraints: usize,
-    ) -> Self {
+    ) -> Result<Self, Error> {
+        let domain = EvaluationDomain::new(verifier_key.n)?;
+        let public_input_roots = public_input_indexes
+            .iter()
+            .map(|index| domain.group_gen_inv.pow(&[*index as u64, 0, 0, 0]))
+            .collect();
         let transcript =
             Transcript::base(label.as_slice(), &verifier_key, constraints);
 
-        Self {
+        Ok(Self {
             label,
             verifier_key,
             opening_key,
             public_input_indexes,
+            public_input_roots,
+            domain,
             transcript,
             size,
             constraints,
-        }
+        })
     }
 
     fn prepare_serialize(
@@ -181,14 +191,14 @@ impl Verifier {
             .map(|n| n as usize)
             .collect();
 
-        Ok(Self::new(
+        Self::new(
             label,
             verifier_key,
             opening_key,
             public_input_indexes,
             size,
             constraints,
-        ))
+        )
     }
 
     /// Verify a generated proof using the current (latest) verification
@@ -226,14 +236,16 @@ impl Verifier {
                 &self.verifier_key,
                 &mut transcript,
                 &self.opening_key,
-                &self.public_input_indexes,
+                &self.domain,
+                &self.public_input_roots,
                 public_inputs,
             ),
             PlonkVersion::V2 | PlonkVersion::V3 => proof.verify(
                 &self.verifier_key,
                 &mut transcript,
                 &self.opening_key,
-                &self.public_input_indexes,
+                &self.domain,
+                &self.public_input_roots,
                 public_inputs,
             ),
         }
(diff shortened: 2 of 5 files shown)


src/proof_system/proof.rs
```diff
@@ -1004,14 +996,56 @@ pub(crate) mod alloc {
...
+            .iter()
+            .filter(|evaluation| *evaluation != &BlsScalar::zero())
+            .map(|evaluation| {
+                let inverse = inverses.next().expect("one inverse per input");
+                *inverse * *evaluation
+            })
+            .sum::<BlsScalar>()
+            * z_h_eval
+            * domain.size_inv;
+        debug_assert!(inverses.next().is_none());
```
moCello: **CAUTION:** `expect` on the verification hot path re-introduces a panic site; two independent nonzero filters must stay in lockstep by hand.

**Context:** The PR removes the old `denom.invert().unwrap()`, but the fused routine relies on the filter that built `denominators` and this second filter over `evaluations` producing the same count so `inverses.next()` never runs dry, then guards that with `expect` + `debug_assert`. Correct today; a future edit to either filter turns a verification error into a process abort.

**Fix:** Pair each inverse with its evaluation by construction: collect the nonzero `(root, evaluation)` pairs once (as the old sparse routine did), build `denominators` from that vector, and zip the pairs with `denominators[1..]` for the sum. That deletes the `expect`, the `debug_assert`, and the second filter.

src/proof_system/proof.rs
```diff
@@ -1004,14 +996,56 @@ pub(crate) mod alloc {
...
-        let denom = n_fr * (z_challenge - BlsScalar::one());
-        z_h_eval * denom.invert().unwrap()
+        domain: &EvaluationDomain,
+    ) -> Result<(BlsScalar, BlsScalar), Error> {
+        if public_input_roots.len() != evaluations.len() {
+            return Err(Error::InconsistentPublicInputsLen {
+                expected: public_input_roots.len(),
+                provided: evaluations.len(),
+            });
+        }
```
moCello: **NOTE:** Length check duplicates the one in `Verifier::verify_with_version` and no test reaches it.

**Context:** Both callers (`Proof::verify`, `Proof::verify_legacy`) are only reached through `verify_with_version`, which already returns `InconsistentPublicInputsLen` on mismatch; `public_input_roots` and `public_input_indexes` have equal length by construction in `Verifier::new`.

**Fix:** Either drop the check (the helper is private and its callers guarantee the invariant) or, if kept as defense in depth, add a unit test next to `fused_evaluation_rejects_a_domain_point` that hits it.

CHANGELOG.md
```diff
@@ -34,6 +34,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remove the `Copy` derive from `Error` [#884]
 - Speed up proof verification by grouping the final pairing-input MSM terms
   and batch-normalizing its G1 inputs [#920]
+- Cache verifier domains and public-input roots, batch their denominator
+  inversions during proof verification, and reject unrepresentable evaluation
+  domain sizes during construction [#927]
```
moCello: **NOTE:** Entry describes internals and bundles three facts; the consumer-visible surface is "verification is faster" and "`Verifier::try_from_bytes` now rejects unrepresentable domain sizes".

**Context:** Deserialization behaviour changed observably: a verifier blob whose `n` cannot be padded to a power of two previously decoded and failed later in `verify`; it now fails in `try_from_bytes` with `InvalidEvalDomainSize`. Cached fields and batched inversion are not visible to a caller.

**Fix:** Split into two entries, e.g. "Speed up proof verification by caching the verifier's evaluation domain and public-input roots [#927]" and "Reject unrepresentable evaluation domain sizes in verifier construction and deserialization [#927]".

src/proof_system/proof.rs
```diff
@@ -1068,6 +1102,7 @@ pub(crate) mod alloc {
         result * numerator
     }
 
+    #[cfg(test)]
     pub(crate) fn compute_barycentric_eval_sparse(
```
moCello: **NOTE:** Test-only reference formula stays in the production module with `pub(crate)` visibility it no longer needs.

**Fix:** Move `compute_barycentric_eval_sparse` into the `tests` submodule below (or make it private); its only user is `fused_lagrange_and_sparse_evaluation_matches_separate_formulas`, which reaches private items via `use super::*`.

src/compiler/verifier.rs
```diff
@@ -273,4 +305,29 @@ mod tests {
...
+        bytes[verifier_key_offset..verifier_key_offset + u64::SIZE].fill(0xff);
+
+        let result =
+            std::panic::catch_unwind(|| Verifier::try_from_bytes(&bytes));
+        assert!(result.is_ok(), "deserializer should never panic");
+        assert!(matches!(
+            result.expect("checked above"),
+            Err(Error::InvalidEvalDomainSize { .. })
+        ));
+    }
```
moCello, line 332: **NOTE:** This test drives the public API only (compile, serialize, mutate, `try_from_bytes`); the crate already hosts that shape in `tests/opening_key_validation.rs`.

**Context:** `tests/opening_key_validation.rs` compiles `SumCircuit`, serializes the verifier and asserts `Verifier::try_from_bytes` rejects mutated bytes, with `HEADER_SIZE` and `read_be_u64` helpers for the offsets this test hardcodes as `48`. Keeping the new case there avoids a second circuit fixture and a `Compiler`/`PublicParameters` dependency inside the `verifier` module's unit tests.

**Fix:** Move the oversized-`n` case next to the malformed-opening-key cases in `tests/opening_key_validation.rs` (or a sibling `verifier_validation.rs`), reusing its fixture and offset helpers.

moCello (commented):
COMMENT — The caching and fused inversion are algebraically identical to the old formulas, the wire format is untouched, and nothing blocks merge; one caution asks for a change on this branch, four notes are yours to take or leave.
- The `expect` inside `compute_lagrange_and_barycentric_evaluations` swaps one panic site for another on the verifier hot path; zip the inverses with the nonzero inputs by construction so no counting invariant is needed.
- Commit `853f843` is subject-only. `33357f5` and `1f67564` each carry a body stating what changed and why, closing with `Resolves #N` — worth matching before merge. The subject also says "evaluations" where the cached values are roots.
- The changelog line should split the perf change from the new deserialization rejection so a consumer sees each on its own.

xevisalle (approved):

xevisalle (approved):

src/compiler/verifier.rs
```diff
@@ -273,4 +305,29 @@ mod tests {
...
+        let verifier_key_offset = 48 + label_len;
+        bytes[verifier_key_offset..verifier_key_offset + u64::SIZE].fill(0xff);
+
+        let result =
+            std::panic::catch_unwind(|| Verifier::try_from_bytes(&bytes));
+        assert!(result.is_ok(), "deserializer should never panic");
+        assert!(matches!(
+            result.expect("checked above"),
+            Err(Error::InvalidEvalDomainSize { .. })
+        ));
```
moCello, line 331: **NOTE:** The no-unwind assertion only bites in debug builds; under `--release` it passes with or without the `checked_next_power_of_two` change, and no CI job runs tests in debug.

**Context:** In release (`overflow-checks = false`), `usize::MAX.next_power_of_two()` wraps to `0`, `0u64.trailing_zeros()` is 64 ≥ `TWO_ADACITY`, so `EvaluationDomain::new` already returned `InvalidEvalDomainSize` before this PR; only debug builds unwound. The rejection assertion is live in release (it covers moving domain construction into `try_from_bytes`); the `catch_unwind` half is the debug-only part. `make test` and the coverage job both run `--release`.

**Fix:** State in a comment on the test that the no-unwind check is a debug-profile regression, or add a debug-profile run of it to CI; either makes the guarantee the test claims match what it exercises. Non-blocking; yours to take or leave.

moCello (approved):
APPROVE — The fused routine pairs each inverse with its input by construction, the changelog is split per fact, and the commit carries a body; nothing asks for a change on this branch beyond one non-blocking note on the oversized-domain test's no-unwind check being exercised only in debug builds.
- Batched formula re-derived against the separate L1 and sparse public-input expressions; equivalent, and the singular-point guard turns the old inversion panic into `ProofVerificationError`.
- Serialization bytes unchanged; caches rebuilt in `try_from_bytes`, unrepresentable `n` now rejected there.

PR #929, opened 2026-08-16
Cover positive curve-addition pole

### LINKED ISSUE #916: Add a regression for the positive curve-addition pole
## Summary

The curve-addition soundness regressions cover the skew = -1 pole, where the x-coordinate coefficient vanishes. The corresponding skew = 1 pole remains established algebraically in the module documentation but has no direct regression fixture.

At this pole, the y-coordinate coefficient vanishes and leaves the assigned y-coordinate free. An isolated fixture should demonstrate that the remaining collapsed numerator still prevents a malformed assignment from satisfying the gate.

## Expected behavior

- Construct off-curve addends whose honest helper wire produces skew = 1.
- Keep the helper wire equal to the actual x1 * y2 product.
- Assign the y-coordinate freed by the pole independently.
- Solve the remaining binding x-coordinate equation.
- Require proving to return Error::CircuitUnsatisfied because the collapsed y-coordinate numerator remains nonzero.
- Keep production circuit behavior and layout unchanged.

## Validation

- Pin the positive-pole condition and all three unweighted residuals explicitly.
- Tie the isolated residual to the production curve-addition widget at two challenges with distinct squares.
- Show that changing the freed y-coordinate does not change the production identity at the pole.
- Add a selector-disabled control proving that the same witness assignment is otherwise accepted.
- Retain an off-curve, non-pole positive control so rejection cannot be attributed to off-curveness alone.

HDauven:
## What changed

- add a nondegenerate honest-helper `skew = +1` curve-addition pole fixture
- pin the helper, pole, binding x-coordinate equation, and all three raw residuals
- compare the isolated y-coordinate residual with the production widget at two separation challenges
- perturb the surviving x-coordinate check and require the production identity to move
- prove that the freed y-coordinate does not affect the identity at the pole
- exercise a shared off-curve non-pole control through the enabled gate and a selector-disabled control for the exact witness
- document the distinct negative- and positive-pole completeness arguments

## Why

The existing regression covers the `skew = -1` pole where the x-coordinate coefficient vanishes. The mirrored positive pole leaves the assigned y-coordinate free, but its collapsed nonzero numerator must still make the circuit unsatisfied.

This is test-only and does not change production code, circuit layouts, transcripts, proofs, or serialization.

## Validation

- `make test`
- `make no-std`
- `make clippy`
- `cargo +nightly fmt --all -- --check`
- focused positive-pole regression
- mutation: removing the production y-coordinate residual makes the regression fail

Resolves #916

### DIFF
--- a/src/composer/tests/soundness/point.rs
+++ b/src/composer/tests/soundness/point.rs
@@ -26,18 +26,20 @@
 //! check collapses to `x1_y2 + y1_x2 = 0` and leaves `x_3` free. The
 //! helper-wire forgeries below avoid both poles — `forced_output` inverts both
 //! denominators — but the malformed scalar-multiplication fixture constructs
-//! off-curve addends whose extended sum has `Z = 0`. The isolated pole
-//! regressions emit only `add_point_gates` for that pair; disabling its sole
-//! curve-addition selector makes the malformed fallback assignment prove,
-//! pinning rejection to the pole gate. A focused witness keeps the helper wire
-//! honest, chooses the x-coordinate the pole frees independently, and solves
-//! the still-binding y-coordinate check, leaving only the collapsed nonzero
-//! x-coordinate numerator. Neither pole is an escape hatch: with `a = -1` a
-//! square and `d` a non-square (asserted in [`subgroup_parameters_hold`]) the
+//! off-curve addends whose extended sum has `Z = 0`. The negative-pole
+//! regressions emit only `add_point_gates` for that pair, while the
+//! positive-pole regression constructs its own honest-wire pole pair.
+//! Disabling the sole curve-addition selector makes each malformed assignment
+//! prove, pinning rejection to the pole gate. Focused witnesses keep the helper
+//! wire honest, choose the coordinate each pole frees independently, and solve
+//! the other coordinate's still-binding check, leaving only the corresponding
+//! collapsed nonzero numerator. Neither pole is an escape hatch: with `a = -1`
+//! a square and `d` a non-square (asserted in [`subgroup_parameters_hold`]) the
 //! addition law is complete, so no on-curve addends put the *honest* wire on a
-//! pole. For off-curve addends at an honest-wire pole, the collapsed output
-//! numerator cannot also vanish without making `d` a square; a helper wire
-//! forged to a pole instead carries a nonzero binding residual.
+//! pole. At the negative pole, a vanishing collapsed numerator would make `d`
+//! a square directly. At the positive pole it would make `d` equal to `-1`
+//! times a square, which is also a square because `-1` is a square. A helper
+//! wire forged to either pole instead carries a nonzero binding residual.
 //!
 //! So without that term a prover picks the wire freely and still satisfies
 //! every other constraint — including, as the steering case shows, picking it
@@ -45,12 +47,12 @@
 //! the attacker's choosing. Those forgeries reproduce the production emission
 //! with only the wire falsified and the coordinates its own checks force;
 //! `forced_output` does that solving, so it is itself pinned against
-//! `compute_quotient_i`. The focused pole fixture separately pins its collapsed
-//! residual against the same production identity. Those pins are this module's
-//! only reaches into `proof_system::widget`, and are deliberate: they tie the
-//! local residual constructions to the widget's own identity, while building
-//! the `ProverKey` by struct literal makes a new field on it break the build
-//! here rather than silently weaken either pin.
+//! `compute_quotient_i`. The focused pole fixtures separately pin their
+//! collapsed residuals against the same production identity. Those pins are
+//! this module's only reaches into `proof_system::widget`, and are deliberate:
+//! they tie the local residual constructions to the widget's own identity,
+//! while building the `ProverKey` by struct literal makes a new field on it
+//! break the build here rather than silently weaken either pin.
 
 use dusk_bls12_381::BlsScalar;
 use dusk_jubjub::{
@@ -382,6 +384,20 @@ fn z_vanishing_addends() -> (JubJubAffine, JubJubAffine) {
     (a, b)
 }
 
+/// Off-curve addends whose honest helper wire makes the curve-addition skew
+/// `+1`, so the production y-coordinate check loses its `y_3` term.
+fn positive_pole_addends() -> (JubJubAffine, JubJubAffine) {
+    let x_1 = BlsScalar::from(2u64);
+    let y_1 = BlsScalar::from(3u64);
+    let x_2 = BlsScalar::from(5u64);
+    let y_2 = invert(BlsScalar::from(30u64) * EDWARDS_D);
+
+    (
+        JubJubAffine::from_raw_unchecked(x_1, y_1),
+        JubJubAffine::from_raw_unchecked(x_2, y_2),
+    )
+}
+
 /// The three unweighted residuals in the production curve-addition identity:
 /// helper consistency, x-coordinate consistency, y-coordinate consistency.
 fn curve_addition_residuals(
@@ -415,6 +431,20 @@ fn output_with_solved_y3(
     let skew = EDWARDS_D * x1_y2 * a.get_v() * b.get_u();
     let y_3 = (a.get_v() * b.get_v() + a.get_u() * b.get_u())
         * invert(BlsScalar::one() - skew);
+
+    JubJubAffine::from_raw_unchecked(x_3, y_3)
+}
+
+/// Solve only the x-coordinate output check. This remains defined at the
+/// positive pole, where the y-coordinate coefficient vanishes.
+fn output_with_solved_x3(
+    a: JubJubAffine,
+    b: JubJubAffine,
+    x1_y2: BlsScalar,
+    y_3: BlsScalar,
+) -> JubJubAffine {
+    let skew = EDWARDS_D * x1_y2 * a.get_v() * b.get_u();
+    let x_3 = (x1_y2 + a.get_v() * b.get_u()) * invert(BlsScalar::one() + skew);
 
     JubJubAffine::from_raw_unchecked(x_3, y_3)
 }
@@ -526,6 +556,15 @@ impl<const ENABLED: bool> IsolatedCurveAdditionCircuit<ENABLED> {
 
         Self::with_assigned_witnesses(a, b, x1_y2, output)
     }
+
+    fn positive_pole_with_solved_x3() -> Self {
+        let (a, b) = positive_pole_addends();
+        let x1_y2 = a.get_u() * b.get_v();
+        let output =
+            output_with_solved_x3(a, b, x1_y2, steering_target().get_v());
+
+        Self::with_assigned_witnesses(a, b, x1_y2, output)
+    }
 }
 
 impl<const ENABLED: bool> Circuit for IsolatedCurveAdditionCircuit<ENABLED> {
@@ -555,6 +594,42 @@ impl<const ENABLED: bool> Circuit for IsolatedCurveAdditionCircuit<ENABLED> {
 
         Ok(())
     }
+}
+
+fn assert_off_curve_non_pole_control(
+    prover: &Prover,
+    verifier: &Verifier,
+    rng: &mut StdRng,
+) {
+    let a = JubJubAffine::from_raw_unchecked(
+        BlsScalar::from(2u64),
+        BlsScalar::from(3u64),
+    );
+    let b = JubJubAffine::from_raw_unchecked(
+        BlsScalar::from(5u64),
+        BlsScalar::from(7u64),
+    );
+    let x1_y2 = a.get_u() * b.get_v();
+    let (x_3, y_3) =
+        forced_output(a.get_u(), a.get_v(), b.get_u(), b.get_v(), x1_y2);
+    let output = JubJubAffine::from_raw_unchecked(x_3, y_3);
+    let control = IsolatedCurveAdditionCircuit::<true>::with_assigned_witnesses(
+        a, b, x1_y2, output,
+    );
+
+    assert!(
+        !bool::from(a.is_on_curve()) && !bool::from(b.is_on_curve()),
+        "the positive control must use off-curve addends",
+    );
+    let skew = EDWARDS_D * x1_y2 * a.get_v() * b.get_u();
+    assert_ne!(skew, BlsScalar::one());
+    assert_ne!(skew, -BlsScalar::one());
+    assert_eq!(
+        curve_addition_residuals(a, b, x1_y2, output),
+        [BlsScalar::zero(); 3],
+        "every raw-gate residual must vanish away from the pole",
+    );
+    assert_verifies(prover, verifier, rng, &control);
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


src/composer/tests/soundness/point.rs
```diff
@@ -382,6 +382,18 @@ fn z_vanishing_addends() -> (JubJubAffine, JubJubAffine) {
...
+/// `+1`, so the production y-coordinate check loses its `y_3` term.
+fn positive_pole_addends() -> (JubJubAffine, JubJubAffine) {
+    let one = BlsScalar::one();
+    let d_inv = invert(EDWARDS_D);
+
+    (
+        JubJubAffine::from_raw_unchecked(one, one),
+        JubJubAffine::from_raw_unchecked(one, d_inv),
+    )
+}
```
moCello, line 399: **CAUTION:** three of the four coordinates are `1`, which collapses the two output numerators onto one field element and leaves several of the fixture's pins unable to fail.

**Context:** `x1·y2 + y1·x2` and `y1·y2 + x1·x2` are both `1/d + 1` at this pair. So `x3 must solve the production check with its exact sign`, `the y3 residual must collapse to its numerator`, and the two-challenge `compute_quotient_i` comparison all hold unchanged for a widget or helper that used one numerator where the other belongs.

`x_1 = 1` costs a second pin: `the helper wire must carry the honest x1*y2 product` reduces to `x1_y2 == y_2`, which a wire assigned `y_2` alone satisfies. `y_1 = x_2 = 1` does the same to the `skew` product.

**Impact:** the rejection itself and the `residuals[2] != 0` pin are unaffected. What weakens is the set of assertions that make this a fixture rather than a bare rejection test.

**Fix:** pick a pole pair with no coordinate equal to `1` and with `x_1 ≠ y_1`, `x_2 ≠ y_2` — choose `x_1`, `y_1`, `x_2` freely and set `y_2 = 1/(D·x_1·y_1·x_2)`, which gives `skew = 1` by construction. `a = (2, 3)`, `b = (5, 1/(30·D))` is one such pair. The fixture's existing off-curve and pole assertions validate the choice.

moCello (commented):
COMMENT — the fixture reaches the positive pole as claimed, rejection is pinned to the curve-addition gate, and the `κ²·χ` weight matches `compute_quotient_i`. What needs work is how much the fixture's assertions can distinguish, one control the test does not carry, and two sentences of the module documentation.

Nothing gates the merge. Five findings below; the two notes ride the same push as the three that carry weight:

- The addend pair's `x_1 = y_1` makes the x- and y-numerators the same field element, so the assertions naming each of them — and the widget comparison — hold equally with the two exchanged. `x_1 = 1` also reduces the helper-wire pin to `x1_y2 == y_2`.
- The off-curve non-pole control that #916 asks for, and that the PR description records as retained, is not in this test — it is in the negative-pole one.
- The module doc still says the isolated pole regressions run on the malformed scalar-multiplication pair; the new fixture builds its own.

One point with no code to anchor: the commit is subject-only. The two it continues, `1f67564` and `38cfced`, each carry a body stating what the fixture assigns, what it pins and that production behaviour is unchanged, closing with `Resolves #N`. The PR description does not become part of the history.

src/composer/tests/soundness/point.rs
```diff
@@ -29,11 +29,11 @@
 //! off-curve addends whose extended sum has `Z = 0`. The isolated pole
 //! regressions emit only `add_point_gates` for that pair; disabling its sole
 //! curve-addition selector makes the malformed fallback assignment prove,
-//! pinning rejection to the pole gate. A focused witness keeps the helper wire
-//! honest, chooses the x-coordinate the pole frees independently, and solves
-//! the still-binding y-coordinate check, leaving only the collapsed nonzero
-//! x-coordinate numerator. Neither pole is an escape hatch: with `a = -1` a
-//! square and `d` a non-square (asserted in [`subgroup_parameters_hold`]) the
+//! pinning rejection to the pole gate. Focused witnesses keep the helper wire
+//! honest, choose the coordinate each pole frees independently, and solve the
```
moCello: **CAUTION:** the doc still routes every isolated pole regression through the malformed scalar-multiplication pair, which the new fixture does not use.

**Context:** `positive_pole_addends` builds `(1, 1)` and `(1, 1/d)`, unrelated to `z_vanishing_addends` — the pair "that pair" names. The following sentence was pluralized for both poles; this one was not. Line 48 drifts the other way: "The focused pole fixture separately pins its collapsed residual" is now two fixtures.

**Impact:** this paragraph is where the pole soundness argument is recorded, and issue #916 cites it as the standing statement of the `skew = 1` case.

**Fix:** scope the "for that pair" sentence to the negative-pole regressions, state that the positive-pole regression builds its own honest-wire pole pair, and pluralize "fixture" on line 48.

src/composer/tests/soundness/point.rs
```diff
@@ -824,6 +859,167 @@ fn collapsed_curve_addition_pole_residual_rejected() {
...
+        challenges[0].square(),
+        challenges[1].square(),
+        "the two challenges must weight the output residuals differently",
+    );
+    for challenge in challenges {
+        let expected = residuals[2] * challenge.square().square() * challenge;
+        for (case, y_3) in [
+            ("focused output", output.get_v()),
+            ("perturbed y3", moved_y3.get_v()),
+        ] {
```
moCello: **NOTE:** the widget pin never shows the surviving `x3` check is present — only the freed coordinate is perturbed, and invariance is what a deleted check would also produce.

**Context:** `residuals[0]` and `residuals[1]` both vanish at this witness, so `expected` matches `compute_quotient_i` identically for a widget whose `xy_consistency` and `x3_consistency` terms were deleted outright. `forced_output_satisfies_the_widget_output_checks` defends against exactly this by perturbing each coordinate and requiring the identity to move, but it cannot reach the poles: `forced_output` inverts both `1 + skew` and `1 - skew`.

The assertion message just above also carries over. It originates in `forced_output_satisfies_the_widget_output_checks`, where both output residuals are solved to zero and two challenges are needed to force both. In either pole test exactly one output residual is nonzero, so the second challenge adds redundancy on the surviving term's `κ` exponent instead.

**Fix:** assert that `compute_quotient_i` with `x_3 + 1` differs from `expected`, mirroring `forced_output_satisfies_the_widget_output_checks`, and reword the challenge assertion to what it pins here. `collapsed_curve_addition_pole_residual_rejected` has the same gap mirrored on `y3`; that one is a separate follow-up.

src/composer/tests/soundness/point.rs
```diff
@@ -29,11 +29,11 @@
...
-//! x-coordinate numerator. Neither pole is an escape hatch: with `a = -1` a
-//! square and `d` a non-square (asserted in [`subgroup_parameters_hold`]) the
+//! pinning rejection to the pole gate. Focused witnesses keep the helper wire
+//! honest, choose the coordinate each pole frees independently, and solve the
+//! other coordinate's still-binding check, leaving only the corresponding
+//! collapsed nonzero numerator. Neither pole is an escape hatch: with `a = -1`
+//! a square and `d` a non-square (asserted in [`subgroup_parameters_hold`]) the
 //! addition law is complete, so no on-curve addends put the *honest* wire on a
 //! pole. For off-curve addends at an honest-wire pole, the collapsed output
 //! numerator cannot also vanish without making `d` a square; a helper wire
```
moCello: **NOTE:** the "collapsed numerator cannot also vanish" argument closes at the positive pole only because `-1` is a square — a step the paragraph does not take. The sentence predates the change; the new regression is what puts weight on it.

**Context:** the two derivations are not symmetric.

At `skew = -1`: `x1·y2 + y1·x2 = 0` with `d·(x1·y2)(y1·x2) = -1` gives `d = (1/(y1·x2))²` directly — `d` a non-square is the whole contradiction.

At `skew = +1`: `y1·y2 + x1·x2 = 0` with `d·x1·x2·y1·y2 = 1` gives `d = -(1/(y1·y2))²`, a square only because `-1` is one. The paragraph cites `a = -1` being a square for completeness of the addition law, not for this.

**Impact:** `subgroup_parameters_hold` asserts both facts, so the argument is sound; as written the positive branch does not reproduce, and a parameter change making `-1` a non-square would not read as touching it.

**Fix:** split the sentence per pole, or state that the positive branch forces `d` into `-1` times a square and so leans on `a = -1` being a square as well as on `d` being a non-square.

src/composer/tests/soundness/point.rs
```diff
@@ -824,6 +859,167 @@ fn collapsed_curve_addition_pole_residual_rejected() {
...
+    let (prover, verifier) = Compiler::compile::<
+        IsolatedCurveAdditionCircuit<true>,
+    >(&pp, b"positive-curve-addition-pole")
+    .expect("compile positive-pole curve addition");
+    assert_verifies(&prover, &verifier, &mut rng, &accepted);
+
+    // The existing negative-pole regression retains an off-curve non-pole
+    // assignment that verifies through this same gate. Disabling the selector
+    // here additionally proves that this exact positive-pole witness is
+    // rejected by no surrounding constraint.
```
moCello: **CAUTION:** this test has no off-curve non-pole control of its own; the comment borrows one from another test, with nothing enforcing the dependency.

**Context:** the "rejection is not off-curveness alone" argument rests on the `off_curve_control` block inside `collapsed_curve_addition_pole_residual_rejected`. Deleting or narrowing that block leaves this test green and this comment silently false; the two tests share no code. The selector-disabled control does not close the same gap — it shows the witness survives with the gate off, not that the enabled gate rejects this pair for sitting at a pole rather than for being off-curve. Only an off-curve non-pole pair proving through the enabled gate separates those.

**Impact:** issue #916 lists "retain an off-curve, non-pole positive control so rejection cannot be attributed to off-curveness alone" among its validation items, and the PR description records it as done — "retain the off-curve non-pole control and add a selector-disabled control for the exact witness". Only the second half holds here.

**Fix:** add the control, mirroring `off_curve_control` in the sibling test: an off-curve non-pole pair through `forced_output`, all three residuals asserted zero, then `assert_verifies`. If it is meant to stay shared instead, name the sibling test in this comment so a future edit has something to grep for, and correct the PR description.

src/composer/tests/soundness/point.rs
HDauven: Changed in 4bd19776: the pole pair is now `(2, 3)` and `(5, 1/(30·D))`, so no coordinate is one and the helper, skew, and two output numerators remain distinguishable.

src/composer/tests/soundness/point.rs
HDauven: Updated in 4bd19776: the module docs now distinguish the negative fixture from the independently constructed positive fixture and pluralize the shared pinning statement.

src/composer/tests/soundness/point.rs
HDauven: Added in 4bd19776: each challenge now also evaluates `compute_quotient_i` with `x3 + 1` and requires it to differ, explicitly pinning the surviving x-coordinate check. The challenge wording now names the surviving y3 weight.

src/composer/tests/soundness/point.rs
HDauven: Clarified in 4bd19776: the negative-pole contradiction makes `d` a square directly, while the positive-pole contradiction makes it `-1` times a square and therefore uses both parameter facts.

src/composer/tests/soundness/point.rs
HDauven: Addressed in 4bd19776: the enabled off-curve non-pole control is now shared scaffolding invoked explicitly by both pole tests, while the exact positive-pole witness retains its selector-disabled control.

xevisalle (approved):

src/composer/tests/soundness/point.rs
```diff
@@ -382,6 +384,20 @@ fn z_vanishing_addends() -> (JubJubAffine, JubJubAffine) {
     (a, b)
 }
 
+/// Off-curve addends whose honest helper wire makes the curve-addition skew
+/// `+1`, so the production y-coordinate check loses its `y_3` term.
+fn positive_pole_addends() -> (JubJubAffine, JubJubAffine) {
+    let x_1 = BlsScalar::from(2u64);
+    let y_1 = BlsScalar::from(3u64);
+    let x_2 = BlsScalar::from(5u64);
+    let y_2 = invert(BlsScalar::from(30u64) * EDWARDS_D);
```
moCello, line 393: **NOTE:** Nothing asserts the fixture's non-degeneracy, so a later edit to these four constants can silently collapse the two output numerators onto one field element again.

**Context:** The pair has to reach `skew = +1` while keeping the numerators `1/(15D) + 15` and `1/(10D) + 10` distinct and each nonzero — that is what lets the helper pin, the numerator pin and the widget comparison fail independently of one another. All three properties follow from the four constants alone, and nothing checks them. The previous fixture failed on exactly this point, with three of its four coordinates equal to `1`, and catching it took a reading of the algebra rather than a test run.

**Fix:** Assert the three properties next to the existing pins: `skew == 1`, the two numerators distinct, and each nonzero.

src/composer/tests/soundness/point.rs
```diff
@@ -745,51 +820,7 @@ fn collapsed_curve_addition_pole_residual_rejected() {
...
-            off_curve_a,
-            off_curve_b,
-            off_curve_wire,
-            off_curve_output,
-        ),
-        [BlsScalar::zero(); 3],
-        "every raw-gate residual must vanish away from the pole",
-    );
-    assert_verifies(&prover, &verifier, &mut rng, &off_curve_control);
+    assert_off_curve_non_pole_control(&prover, &verifier, &mut rng);
```
moCello, line 839: **NOTE:** The negative-pole test has no widget-level perturbation pinning its surviving `y3` check, so it proves less than the positive-pole test added here.

**Context:** At the negative pole `1 + skew = 0` frees `x_3` while `1 - skew = 2` binds `y_3`. The test perturbs `x_3` (point.rs:767, :791) and shows the residual does not move, which establishes the freed side; nothing evaluates `compute_quotient_i` against a moved `y_3` to show the bound check is still present. The positive-pole test does exactly that for its own surviving coordinate at :976-990. `forced_output_satisfies_the_widget_output_checks` perturbs both coordinates away from either pole, where both checks are active regardless, so it does not reach the pole-specific claim.

**Fix:** Mirror the `perturbed_x3` block into `collapsed_curve_addition_pole_residual_rejected` with `y_3 + 1`, asserting the identity moves.

moCello (approved):
APPROVE — every finding from the previous round is addressed at the code, and re-deriving the fixture's algebra against the widget and against dusk-jubjub turned up nothing new. Nothing gates; the two notes below are yours to take or leave, and both are small enough to ride one push.

The replacement pair `(2, 3)`, `(5, 1/(30·D))` reaches `skew = 1` with two distinct, individually nonzero output numerators, so the helper pin, the numerator pin, and the widget comparison can each fail on its own. It differs from the off-curve non-pole control in `y_2` alone, which makes that control a tighter separation of pole-ness from off-curveness than it was. The `x_3 + 1` perturbation shifts the production identity by `-2κχ`, which is what turns the widget comparison into evidence that the surviving x-coordinate check is present rather than absent. The `Z = 0` assertion reproduces from the extended addition formula, and the module doc's positive-pole branch — `d` forced to `-1` times a square — derives as written.

- What the fixture rests on is not asserted anywhere: distinct output numerators and no coordinate equal to one both follow from the four constants alone. That is the property the last round had to catch by eye, so it is worth closing here rather than later.
- The negative-pole test still proves less than its new positive-pole counterpart — it pins the freed coordinate but not the surviving one. Both notes are inline.

PR #932, opened 2026-08-17
Reject invalid compressed-circuit indices

### LINKED ISSUE #930: Reject invalid compressed-circuit indices without panicking
## Summary

`CompressedCircuit::from_bytes` reconstructs constraint witness indices from serialized data and passes them to the composer without first checking that they refer to allocated witnesses. An out-of-range index reaches the permutation map's internal assertion, so `Compiler::compile_with_compressed` can panic instead of returning `Error::InvalidCompressedCircuit`.

The same boundary should validate the remaining index relationships and public-input ordering before reconstruction begins.

Malformed compressed-circuit artifacts can panic consumers that expect the checked decoding API to return an error. This primarily affects applications that accept compressed circuit descriptions from untrusted or corrupt storage.

Selector-scalar and polynomial lookups already return `InvalidCompressedCircuit` when out of range. Witness indices currently do not receive the same treatment, and malformed public-input indices can be silently ignored.

## Acceptance criteria

- Every constraint witness index is required to be smaller than the serialized witness count before any gate is appended.
- Polynomial and scalar references remain checked without panicking.
- Public-input indices are strictly ordered, unique, and within the constraint range.
- Malformed inputs return `Error::InvalidCompressedCircuit` without unwinding.
- Valid compressed circuits remain byte-for-byte compatible and continue to compile, prove, and verify normally.
- Regressions cover each rejected index class and a valid-circuit control.

HDauven:
## Summary

- validate compressed-circuit witness, polynomial and selector indices before rebuilding the composer
- reject duplicate, descending and out-of-range public-input positions
- add no-unwind regressions for every wire and selector slot

## Rationale

Compressed constraints previously converted serialized witness indices directly into `Witness` values. An out-of-range index then reached the permutation map and triggered an assertion instead of returning a decoding error. Public-input positions could also be silently ignored when duplicated, out of order or out of range.

Malformed indices now return `Error::InvalidCompressedCircuit` before composer allocation and mutation. The compressed format and valid circuit behavior are unchanged.

## Validation

- `make test`
- `make clippy`
- `make no-std`
- `cargo check --target thumbv6m-none-eabi --no-default-features --features alloc,rkyv-impl,rkyv/size_32`
- `make doc`
- `make doc-internal`
- `cargo bench --no-run`
- `cargo run --release --example circuit`
- mutation check reproducing the original permutation assertion when the witness-index guard is removed

Resolves #930

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Reject invalid witness, polynomial, selector and public-input indices when
+  deserializing compressed circuits [#930]
 - Reject trailing data when deserializing compressed circuits [#913]
 - Validate cached permutation linear evaluations during checked `Prover`
   deserialization [#908]
@@ -753,6 +755,7 @@ is necessary since `rkyv/validation` was required as a bound.
 - Proof system module.
 
 <!-- ISSUES -->
+[#930]: https://github.com/dusk-network/plonk/issues/930
 [#914]: https://github.com/dusk-network/plonk/issues/914
 [#913]: https://github.com/dusk-network/plonk/issues/913
 [#910]: https://github.com/dusk-network/plonk/issues/910
--- a/src/composer/compress.rs
+++ b/src/composer/compress.rs
@@ -42,6 +42,24 @@ pub struct CompressedPolynomial {
     pub q_variable_group_add: usize,
 }
 
+impl CompressedPolynomial {
+    fn scalar_indices(&self) -> [usize; 11] {
+        [
+            self.q_m,
+            self.q_l,
+            self.q_r,
+            self.q_o,
+            self.q_f,
+            self.q_c,
+            self.q_arith,
+            self.q_range,
+            self.q_logic,
+            self.q_fixed_group_add,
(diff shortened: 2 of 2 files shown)


xevisalle (approved):

Neotamandua (approved):

moCello (approved):
APPROVE — The validation pass is correct: the witness bound matches an empty `Composer::uninitialized()`, strict public-input ordering matches what `from_composer` emits, the wire format is untouched, and every rejected class has a test plus a valid control. Nothing asks for a change on this branch; two non-blocking notes:
- The codecov drop splits two ways: the twelve pre-existing `.get(..).ok_or(InvalidCompressedCircuit)?` arms in `from_bytes` (polynomial + 11 selectors), which `validate_indices` now makes unreachable, and the new `checked_add` overflow arm in `validate_indices` itself, which needs a scalar count near `usize::MAX` to fire. Leaving both as defense in depth is fine; no change needed for this PR.
- Pre-existing, for a follow-up issue: `witnesses` itself is unbounded, so a decoded value near `usize::MAX` drives `(0..witnesses).for_each(append_witness)` into memory exhaustion from a few bytes. No structural bound exists (unreferenced witnesses are legal), so this needs a decision on a cap rather than an index check.

PR #934, opened 2026-08-17
Parallelize final FFT stages

### LINKED ISSUE #933: Parallelize final FFT stages and share butterfly implementation
The parallel FFT added for large standard-library domains distributes independent chunks while `chunk_count >= 4`. The final two butterfly stages and the bit-reversal pass remain serial, and the serial and parallel paths duplicate the same butterfly and bit-reversal algorithms.

For domains around 2^17 to 2^20, the two serial stages account for roughly 10–12% of butterfly work before considering the serial bit reversal. This leaves avoidable serial work and duplicate implementation paths.

## Proposed work

- Parallelize independent sub-ranges within a butterfly chunk once fewer than four whole chunks remain.
- Seed each sub-range with the correct twiddle factor for its starting offset.
- Share bit-reversal and butterfly helpers between the serial and parallel paths.
- Preserve exact FFT output for all supported power-of-two domains.
- Benchmark representative large domains before and after the change.

## Acceptance criteria

- Parallel and serial FFT results remain identical.
- Small domains and `no_std` builds retain the serial path.
- Tests cover the late-stage sub-range twiddle offsets.
- Repository benchmarks demonstrate whether the additional parallelism is worthwhile.

Refs #921

HDauven:
## Summary

- share bit-reversal and butterfly helpers between the serial and parallel FFT paths
- split final stages for domains of at least `2^12` coefficients on Rayon pools with at least four workers
- derive public twiddle seeds with one `pow_vartime(range_len)` step and repeated multiplication, including uneven tails
- compare the nested five-transform quotient path slot-for-slot with independent `domain.coset_fft` results

## Performance

The production-shaped four-input Phoenix proving workload was run on the dedicated CPX42 reference worker (8 vCPU, 16 GiB) with `RAYON_NUM_THREADS=8`. It has 131,024 constraints and a `2^17` proving domain.

This comparison used current `master` at `6e21eff1` and exact candidate `2bc96aec` (tree `1b2fdb05`). Each revision ran in four counterbalanced processes, each with one warm-up and six measured proofs, for 24 measured and verified proofs per revision.

| Revision | Run 1 | Run 2 | Run 3 | Run 4 | Mean median |
| --- | ---: | ---: | ---: | ---: | ---: |
| `master` | 6.9590 s | 6.8645 s | 6.9764 s | 6.8526 s | 6.9131 s |
| Candidate | 6.6582 s | 6.6212 s | 6.7265 s | 6.6852 s | 6.6728 s |

The candidate reduces the mean proof median by 3.48% (240.3 ms) and increases implied throughput by 3.60%. Every paired round favored the candidate, by 2.44% to 4.32%; the paired mean has a 95% t-interval of 2.24% to 4.70%. Across all 24 samples per revision, the raw median reduction is 3.74%.

Mean verification medians and peak RSS were effectively unchanged (3.146 ms versus 3.152 ms; 2,170,086 KiB versus 2,169,854 KiB).

### Final-stage cutoff

Complete FFTs were also benchmarked with only the final-stage policy changed, alternating forced-serial and forced-parallel order. Every measured combination favored final-stage parallelism:

| Domain | Rayon workers | Parallel change |
| --- | --- | ---: |
| `2^12` | 4–8 | -8.42% to -13.33% |
| `2^14` | 4–8 | -13.94% to -26.02% |
| `2^17` | 4–8 | -17.57% to -36.36% |

This supports the `2^12` work-size and four-worker boundary. Smaller domains retain the serial FFT path.

## Validation

- `make fmt`
- `make clippy`
- `make no-std`
- `make test`
- `cargo test --release --all-features`
- serial/parallel equivalence and FFT/IFFT roundtrips at the three-/four-worker boundary and with an uneven nine-way partition
- independent per-slot quotient coset FFT equivalence under one-, eight-, and nine-worker Rayon pools
- mutation controls for slot order, stepped twiddle seed/uneven tail, and work-size cutoff
- production-shaped nested-Rayon Phoenix benchmark with every measured proof verified

Resolves #933

### DIFF
--- a/src/fft/domain.rs
+++ b/src/fft/domain.rs
@@ -366,6 +366,20 @@ pub(crate) mod alloc {
     }
 
     #[cfg(feature = "std")]
+    pub(super) const PARALLEL_FINAL_FFT_MIN_LEN: usize = 1 << 12;
+    #[cfg(feature = "std")]
+    pub(super) const PARALLEL_FINAL_FFT_MIN_THREADS: usize = 4;
+
+    #[cfg(feature = "std")]
+    pub(super) fn should_parallelize_final_fft_stages(
+        len: usize,
+        threads: usize,
+    ) -> bool {
+        len >= PARALLEL_FINAL_FFT_MIN_LEN
+            && threads >= PARALLEL_FINAL_FFT_MIN_THREADS
+    }
+
+    #[cfg(feature = "std")]
     fn best_fft(a: &mut [BlsScalar], omega: BlsScalar, log_n: u32) {
         const PARALLEL_FFT_MIN_LEN: usize = 1 << 12;
         const PARALLEL_FFT_MIN_CHUNKS: usize = 4;
@@ -379,12 +393,7 @@ pub(crate) mod alloc {
         let n_u32 = n as u32;
         assert_eq!(n_u32, 1 << log_n);
 
-        for k in 0..n_u32 {
-            let rk = bitreverse(k, log_n);
-            if k < rk {
-                a.swap(rk as usize, k as usize);
-            }
-        }
+        bitreverse_permute(a, log_n);
 
         let mut m = 1usize;
         for _ in 0..log_n {
@@ -395,6 +404,13 @@ pub(crate) mod alloc {
             if chunk_count >= PARALLEL_FFT_MIN_CHUNKS {
                 a.par_chunks_mut(chunk_len)
                     .for_each(|chunk| butterfly_chunk(chunk, m, w_m));
+            } else if should_parallelize_final_fft_stages(
+                n,
+                rayon::current_num_threads(),
+            ) {
+                for chunk in a.chunks_mut(chunk_len) {
+                    parallel_butterfly_chunk(chunk, m, w_m);
+                }
             } else {
                 for chunk in a.chunks_mut(chunk_len) {
                     butterfly_chunk(chunk, m, w_m);
@@ -415,6 +431,15 @@ pub(crate) mod alloc {
         r
     }
 
+    fn bitreverse_permute(a: &mut [BlsScalar], log_n: u32) {
+        for k in 0..a.len() as u32 {
+            let rk = bitreverse(k, log_n);
+            if k < rk {
+                a.swap(rk as usize, k as usize);
+            }
+        }
+    }
+
     pub(crate) fn serial_fft(
         a: &mut [BlsScalar],
         omega: BlsScalar,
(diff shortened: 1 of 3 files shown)


moCello (commented):
COMMENT — The subrange partition and twiddle seeding are correct and the serial path is behaviorally unchanged; one test regression should be fixed, and two notes are yours to take or leave.
- The rewritten quotient test compares `compute_coset_evaluations` against itself and no longer proves slot order; restoring the `coset_fft` reference is a one-line change.
- Notes: constant-time `pow` per subrange dominates at small parallel domains, and the eight-thread cutoff is unexplained.

src/proof_system/quotient_poly.rs
```diff
@@ -312,20 +312,39 @@ fn compute_permutation_checks(
...
+                    .collect(),
+            )
+        });
+        let reference_pool = rayon::ThreadPoolBuilder::new()
+            .num_threads(1)
+            .build()
+            .unwrap();
+        let expected = reference_pool.install(|| {
+            compute_coset_evaluations(&domain, polynomials.each_ref())
         });
```
moCello, line 336: **CAUTION:** `coset_evaluations_preserve_input_order_across_nested_rayon_pools` no longer asserts input order — expected and actual both come from `compute_coset_evaluations`.

**Context:** The replaced test compared against `polynomials.each_ref().map(|poly| domain.coset_fft(poly))`, an independent per-slot reference. Now a deterministic slot permutation inside `compute_coset_evaluations` (for example the `zip` pairing slot `i` with polynomial `4 - i`) passes on every pool size, since the 1-thread run misorders identically.

**Severity-rationale:** The end-to-end prover tests would still catch a misordered slot, so this is a unit-test claim that no longer holds rather than an open hole; the fix is one line.

**Fix:** Restore the independent reference inside the single-thread pool: `let expected = reference_pool.install(|| polynomials.each_ref().map(|poly| domain.coset_fft(poly)));`. That keeps both properties — slot order against a per-slot reference, and pool invariance against the 8- and 9-thread runs.

src/fft/domain.rs
```diff
@@ -418,54 +429,68 @@ pub(crate) mod alloc {
...
+        right: &mut [BlsScalar],
+        w_m: BlsScalar,
+        offset: usize,
+    ) {
+        debug_assert_eq!(left.len(), right.len());
+        let mut w = if offset == 0 {
+            BlsScalar::one()
+        } else {
+            w_m.pow(&[offset as u64, 0, 0, 0])
+        };
```
moCello: **NOTE:** Each subrange seeds its twiddle with constant-time `pow` — 256 squarings plus 256 multiplications — which at small parallel domains outweighs the subrange's own butterfly work.

**Context:** At the `2^12` cutoff with eight threads a subrange is 128 or 256 butterflies (~256–512 multiplications) against ~512 for the seed, so the final two stages spend two to three times the CPU of the serial path; at `2^17` the seed is 3–6% of a subrange. Wall clock still improves on an idle pool, but under the nested five-transform quotient path all workers are already busy, so CPU cost is what counts. The exponent is public, so constant time buys nothing here; `pow_vartime` alone only drops the multiplications and keeps the 256 squarings. `benches/plonk.rs` already proves at `2^5`–`2^17`; the PR measured only the `2^17` external workload, so the small-domain end is unmeasured.

**Fix:** In `parallel_butterfly_chunk` compute `step = w_m.pow_vartime(range_len)` once and derive the per-subrange seeds by repeated multiplication (one multiplication per subrange) before fanning out, passing the seed instead of `offset` to `butterfly_range`.

src/fft/domain.rs
```diff
@@ -360,6 +360,9 @@ pub(crate) mod alloc {
         serial_fft(a, omega, log_n)
     }
 
+    #[cfg(feature = "std")]
+    pub(super) const PARALLEL_FFT_MIN_THREADS: usize = 8;
```
moCello: **NOTE:** The eight-worker cutoff carries no rationale in code or in #933, and it leaves four- to seven-thread pools with fully serial final stages.

**Context:** The earlier stages already split into `par_chunks_mut` for any pool size once four chunks exist, so a four-thread pool parallelizes a 4-chunk stage but not the 2-chunk stage that follows it, although the subrange there is at least 256 butterflies wide. The PR measured only an eight-thread pool, so the diff cannot show whether the cutoff was chosen from data or as a guard.

**Fix:** Add a doc comment on the constant stating why eight and what measurement backs it; if it is a guard rather than a measured floor, gate on `range_len` instead of thread count.

src/proof_system/quotient_poly.rs
HDauven: Restored the independent one-thread reference as `polynomials.each_ref().map(|poly| domain.coset_fft(poly))`; the nested 8/9-worker result is again checked slot-for-slot against it.

src/fft/domain.rs
HDauven: Replaced per-range constant-time exponentiation with one public `pow_vartime(range_len)` step and repeated multiplication. The uneven nine-way test rejects an off-by-one seed/tail mutation.

src/fft/domain.rs
HDauven: Benchmarked complete FFTs on the CPX42 at `2^12`, `2^14`, and `2^17` with 4–8 workers. All 15 combinations improved, including 8.4–13.3% at `2^12`, so the measured gate is now `>=2^12` and `>=4` workers with boundary mutation coverage.

moCello (approved):
APPROVE — All three earlier points are addressed: the quotient test again checks slot order against an independent single-thread `coset_fft` reference, the per-range twiddle seed is now one public `pow_vartime` step plus repeated multiplication, and the final-stage gate is a measured ≥2^12 / ≥4-worker floor with boundary tests. Sub-range partition and seeding trace identical to the serial loop, including uneven tails; `no_std` keeps the serial path. No new findings.
- Optional, not blocking: a one-line doc comment on `PARALLEL_FINAL_FFT_MIN_LEN`/`PARALLEL_FINAL_FFT_MIN_THREADS` pointing at the benchmark rationale would keep it discoverable from the code.

PR #936, opened 2026-08-17
Reject over-degree prover-key polynomials

### LINKED ISSUE #935: Reject over-degree fixed polynomials in decoded prover keys
## Summary

`ProverKey::from_slice` accepts serialized fixed polynomials whose coefficient count exceeds the circuit domain size `n`.

Canonical compilation interpolates each selector and permutation polynomial from exactly `n` circuit rows, so every resulting fixed polynomial has degree smaller than `n`. A malformed artifact can currently violate that invariant while still passing checked decoding.

An inconsistent prover artifact can be accepted and fail later during proof construction or produce proofs that do not verify. This is local prover-artifact integrity and availability hardening; the verifier does not deserialize or trust these polynomials.

The decoder already validates the shared `8n` evaluation domain and the derived linear and vanishing evaluations, but it does not bound the 15 serialized fixed-polynomial coefficient vectors against `n`.

## Acceptance criteria

- Reject every serialized fixed polynomial whose coefficient count exceeds `n`.
- Apply the bound before reading or allocating the polynomial coefficients.
- Return a decoding error rather than constructing a usable `ProverKey`.
- Cover all selector and permutation polynomial slots with malformed-byte regressions.
- Preserve valid prover-key bytes and round-trip behavior.
- Preserve `no_std` support and proof/verifier behavior for valid artifacts.

HDauven:
## Summary

- bound every serialized fixed-polynomial coefficient vector by the circuit domain size before reading its coefficients
- reject malformed prover keys whose fixed-polynomial degree is at least `n`
- cover all 15 selector and permutation polynomial slots with raw-byte regressions

## Root cause

`ProverKey::from_slice` validated the shared `8n` evaluation domain and derived caches but trusted each serialized polynomial length independently of the circuit size. Canonical compilation interpolates these fixed polynomials from exactly `n` rows, so their degree must be smaller than `n`.

The shared polynomial reader now rejects an announced coefficient count greater than `n` before reading or allocating those coefficients.

## Impact

Malformed or inconsistent local prover artifacts are rejected at the decoding boundary instead of being accepted and failing later during proving or producing unverifiable proofs.

Valid prover-key bytes, proof generation, verification, transcript behavior, and serialization formats are unchanged. The verifier does not deserialize these fixed polynomials.

## Validation

- pre-fix reproduction accepting an over-degree polynomial
- mutation control showing the regression fails without the new bound
- raw-byte rejection coverage for all 15 fixed-polynomial positions
- `make test`
- `make fmt`
- `make clippy`
- `make no-std`

Resolves #935

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Reject unrepresentable evaluation domain sizes during verifier construction
   and checked deserialization [#927]
+- Reject fixed polynomials with degree at least the circuit domain size during
+  checked `ProverKey` deserialization [#935]
 - Reject invalid witness, polynomial, selector and public-input indices when
   deserializing compressed circuits [#930]
 - Reject trailing data when deserializing compressed circuits [#913]
@@ -769,6 +771,7 @@ is necessary since `rkyv/validation` was required as a bound.
 - Proof system module.
 
 <!-- ISSUES -->
+[#935]: https://github.com/dusk-network/plonk/issues/935
 [#930]: https://github.com/dusk-network/plonk/issues/930
 [#925]: https://github.com/dusk-network/plonk/issues/925
 [#921]: https://github.com/dusk-network/plonk/issues/921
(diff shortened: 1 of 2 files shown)


moCello (approved):
APPROVE — The bound matches what compilation produces (`n = domain.size()`, every fixed polynomial interpolated from `n` rows), it runs before any coefficient bytes are read, and the regression covers all 15 slots. Nothing asks for a change.
- Non-blocking, for a later pass if ever touched again: in `poly_from_reader`, a length that does not fit `usize` still maps to `NotEnoughBytes` (pre-existing), while `n` a few lines up maps the same conversion failure to `InvalidData`; a `u64` over `usize` is by definition over the `n` bound.

PR #938, opened 2026-08-19
Normalize serialized prover-key length errors

### LINKED ISSUE #937: Normalize oversized serialized-length errors during prover-key decoding
## Summary

`ProverKey::from_slice` currently classifies equivalent integer-conversion failures differently. The circuit-domain size `n` maps a serialized `u64` that does not fit `usize` to `InvalidData`, while the fixed-polynomial length conversion inside `poly_from_reader` maps the same condition to `NotEnoughBytes`.

An unrepresentable polynomial length is malformed metadata rather than truncated input. It is also necessarily greater than every valid `usize` domain bound. The decoder already rejects it, so this is error-taxonomy cleanup rather than a correctness or security fix.

## Acceptance criteria

- Map an unrepresentable fixed-polynomial coefficient count to `InvalidData`.
- Continue returning `NotEnoughBytes` for genuinely truncated coefficient bytes.
- Cover the distinction where the target width permits the conversion-overflow case, or isolate the conversion policy in a target-independent unit test.
- Preserve the serialized format, valid round trips, proof behavior, and `no_std` support.

Follow-up to #936 and its review.

HDauven:
## Summary

- normalize all serialized prover-key length conversions through one helper
- report unrepresentable metadata lengths as `InvalidData`
- preserve `NotEnoughBytes` for genuinely truncated coefficient data
- pin the conversion policy independently of the host pointer width

## Compatibility

Valid prover-key bytes and round trips are unchanged. This only makes the error classification for malformed, unrepresentable polynomial lengths consistent with the other serialized length fields.

## Validation

- `make test`
- `make clippy`
- `make no-std`
- `make doc`
- `make fmt`
- mutation check: restoring `NotEnoughBytes` for conversion failure makes the new regression fail

Resolves #937

### DIFF
--- a/src/proof_system/widget.rs
+++ b/src/proof_system/widget.rs
@@ -207,6 +207,13 @@ pub(crate) mod alloc {
     use dusk_bls12_381::BlsScalar;
     use merlin::Transcript;
 
+    pub(super) fn serialized_len_from_u64<T>(len: u64) -> Result<T, Error>
+    where
+        T: TryFrom<u64>,
+    {
+        T::try_from(len).map_err(|_| dusk_bytes::Error::InvalidData.into())
+    }
+
     impl VerifierKey {
         fn seed_transcript_inner(
             &self,
@@ -439,11 +446,12 @@ pub(crate) mod alloc {
         /// Deserializes a slice of bytes into a [`ProverKey`].
         pub fn from_slice(bytes: &[u8]) -> Result<ProverKey, Error> {
             let mut buffer = bytes;
-            let n = usize::try_from(u64::from_reader(&mut buffer)?)
-                .map_err(|_| dusk_bytes::Error::InvalidData)?;
-            let evaluations_size =
-                usize::try_from(u64::from_reader(&mut buffer)?)
-                    .map_err(|_| dusk_bytes::Error::InvalidData)?;
+            let n = serialized_len_from_u64::<usize>(u64::from_reader(
+                &mut buffer,
+            )?)?;
+            let evaluations_size = serialized_len_from_u64::<usize>(
+                u64::from_reader(&mut buffer)?,
+            )?;
             let evaluations_domain_size =
                 n.checked_mul(8).ok_or(dusk_bytes::Error::InvalidData)?;
             if evaluations_domain_size.checked_next_power_of_two()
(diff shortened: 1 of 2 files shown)


src/proof_system/widget.rs
```diff
@@ -207,6 +207,13 @@ pub(crate) mod alloc {
     use dusk_bls12_381::BlsScalar;
     use merlin::Transcript;
 
+    pub(super) fn serialized_len_from_u64<T>(len: u64) -> Result<T, Error>
+    where
+        T: TryFrom<u64>,
+    {
+        T::try_from(len).map_err(|_| dusk_bytes::Error::InvalidData.into())
+    }
```
moCello, line 215: **NOTE:** Follow-up issue, nothing to change here: three other serialized-length decoders still read lengths with a truncating `as usize`.

**Context:** `VerifierKey::from_bytes` in this file reads the circuit size as `u64::from_reader(&mut buffer)? as usize`, and `from_polynomial_commitments` stores it unvalidated. `Prover::try_from_bytes` and `Verifier::try_from_bytes` do the same for six header lengths each, plus every public-input index. On a 32-bit target these wrap instead of erroring, so decoding stops being injective — `n` and `n + 2^32` yield the same `VerifierKey`. `Verifier::try_from_bytes` is public and takes arbitrary bytes, and reaches `VerifierKey::from_bytes` through `from_slice`. The divergence is the concrete cost: `EvaluationDomain::new` rejects the oversized `n` on 64-bit, and on 32-bit only ever sees the truncated one. `constraints` in the two `compiler` decoders wraps the same way into `circuit_domain_sep` via `Transcript::base`, and `Verifier` never cross-checks it against `verifier_key.n` on any target width.

**Fix:** Track as an issue and route those reads through this helper. Adopting it at `VerifierKey::from_bytes` first needs the helper to return `Result<T, dusk_bytes::Error>` — that impl is `Serializable` with `type Error = dusk_bytes::Error`, so `?` cannot widen a plonk `Error` there, while the three call sites here would still widen through the existing `From`. The two `compiler` sites additionally need `pub(crate)` and a home `compiler` can reasonably import from; `proof_system::widget::alloc` is not it.

moCello (approved):
APPROVE — the reclassification is correct and complete for the site it targets: the serialized format, valid round trips and 64-bit behavior are unchanged, and the new unit test pins the policy without depending on the host pointer width.

One note below, and its whole remedy is an issue — nothing on this branch needs to change to merge.

Two further points with no inline home, both also issue-shaped:

Truncation is already reported through two distinct variants depending on where it is caught — `Error::NotEnoughBytes` from `split_at_checked`, and `Error::BytesError(dusk_bytes::Error::NotEnoughBytes)` when the buffer ends inside a length header read by `u64::from_reader`. Nothing asserts the second, and a caller matching the first misses it. A PR normalizing decode-error classification is the natural place to notice the split; whether to collapse the two is a maintainer judgment call, and the follow-up issue is the place to record it.

`ProverKey` derives `Archive` with `archive_attr(derive(CheckBytes))`, so the rkyv checked path applies structural validation only. None of the semantic checks `from_slice` performs — the degree bound from #935, the length classification here, the linear and vanishing coset-evaluation checks — run on that path. Worth its own issue, given #903 did this work for `PublicParameters`.

PR #939, opened 2026-08-19
Bound compressed-circuit allocation

### LINKED ISSUE #931: Bound compressed-circuit decompression and allocation
## Summary

`CompressedCircuit::from_bytes` inflates its complete input with an unbounded `decompress_to_vec` call. Serialized collection counts and the witness count can then drive allocation and reconstruction work without a limit tied to the circuit capacity.

A small compressed artifact can therefore cause disproportionate memory or CPU use before decoding returns an error.

This primarily affects consumers that accept compressed circuit descriptions from untrusted or corrupt storage. Rusk currently regenerates its known circuit descriptions locally, but the public library API does not require callers to use a trusted source.

The exact limit policy is compatibility-sensitive: legitimate large circuits must remain supported, and an arbitrary global byte ceiling would be difficult to justify.

## Acceptance criteria

- Inflated output is bounded before an attacker-controlled amount of memory is committed.
- Serialized collection counts are rejected before disproportionate allocation or iteration.
- Limits are derived from the supplied public-parameter or circuit capacity where possible, rather than an unexplained global constant.
- Limit violations return `Error::InvalidCompressedCircuit` without unwinding or aborting.
- Existing valid compressed circuits continue to compile normally.
- Regressions cover a high-ratio compressed payload, disproportionate counts, boundary-valid inputs, and normal circuit compilation.
- The selected limits and compatibility rationale are documented before the implementation is made ready for review.

HDauven:
## Summary

- derive compressed-circuit limits from the supplied public parameters
- bound DEFLATE output before committing unbounded memory
- validate serialized collection counts before allocating their elements
- treat serialized witness indices as labels and allocate only labels referenced by gates

The compressed format remains unchanged. Canonical descriptions produced by `Circuit::compress()`, including descriptions with sparse or unused witness labels, reconstruct byte-identical prover and verifier keys.

## Root cause

`compile_with_compressed` previously decompressed the complete input and allowed serialized collection lengths and the raw witness count to drive allocation before relating the description to the supplied public-parameter capacity.

The decoder now applies the same constraint-capacity boundary that compilation will eventually enforce. The packed-byte ceiling follows the worst-case canonical MessagePack representation. Collection lengths are checked before their elements are allocated, and witness labels are validated against the serialized namespace before being densely remapped by actual gate use. A large or sparse declared witness namespace therefore cannot cause proportional reconstruction work.

## Validation

- full default and all-feature release test suites
- Clippy and formatting
- `no_std` alloc and core-only builds for `thumbv6m-none-eabi`
- malformed inflation, collection-count, sparse-witness-label and boundary regressions
- direct and compressed compilation of a canonical circuit with unused witnesses produce byte-identical prover and verifier keys
- all four retained Rusk circuit descriptions compile with the corresponding CRS

Resolves #931

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -45,6 +45,8 @@ impl PlonkVersion {
 pub struct Compiler;
 
 impl Compiler {
+    const CIRCUIT_SIZE_PADDING: usize = 6;
+
     /// Create a new arguments set from a given circuit instance
     ///
     /// Use the default implementation of the circuit
@@ -80,14 +82,32 @@ impl Compiler {
 
     /// Generates a [Prover] and [Verifier] from a buffer created by
     /// [Circuit::compress].
+    ///
+    /// The supplied public parameters bound decompression and circuit
+    /// reconstruction. Compressed descriptions containing more data than can
+    /// be compiled with `pp` are rejected as malformed.
     pub fn compile_with_compressed(
         pp: &PublicParameters,
         label: &[u8],
         compressed: &[u8],
     ) -> Result<(Prover, Verifier), Error> {
-        let composer = Composer::from_bytes(compressed)?;
+        let max_constraints = Self::max_constraints(pp);
+        let composer = Composer::from_bytes(compressed, max_constraints)?;
 
         Self::compile_with_composer(pp, label, &composer)
+    }
+
+    fn max_constraints(pp: &PublicParameters) -> usize {
+        let available = pp
+            .max_degree()
+            .saturating_sub(PublicParameters::ADDED_BLINDING_DEGREE);
+        let max_domain_size = if available == 0 {
+            0
+        } else {
+            1usize << (usize::BITS - available.leading_zeros() - 1)
+        };
+
+        max_domain_size.saturating_sub(Self::CIRCUIT_SIZE_PADDING)
     }
 
     /// Create a new arguments set from a given circuit instance
@@ -98,7 +118,8 @@ impl Compiler {
         label: &[u8],
         composer: &Composer,
     ) -> Result<(Prover, Verifier), Error> {
-        let n = (composer.constraints() + 6).next_power_of_two();
+        let n = (composer.constraints() + Self::CIRCUIT_SIZE_PADDING)
+            .next_power_of_two();
 
         let (commit, opening) = pp.trim(n)?;
 
--- a/src/composer/compress.rs
+++ b/src/composer/compress.rs
@@ -95,6 +95,15 @@ pub struct CompressedCircuit {
 }
 
 impl CompressedCircuit {
+    // Canonical MessagePack uses at most nine bytes per usize, two per u8,
+    // and five for each vector header. For every constraint, the encoder can
+    // add one public-input index, eleven 32-byte scalars, one polynomial with
+    // eleven indices, and one constraint with five indices.
+    const PACKED_FIXED_BYTES: usize = 30;
+    const PACKED_BYTES_PER_CONSTRAINT: usize = 857;
+
+    const SELECTORS_PER_POLYNOMIAL: usize = 11;
+
     fn validate_indices(&self, base_scalars: usize) -> Result<(), Error> {
         let scalar_count = base_scalars
             .checked_add(self.scalars.len())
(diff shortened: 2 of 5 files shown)


src/composer/compress.rs
```diff
@@ -95,6 +95,15 @@ pub struct CompressedCircuit {
...
 
 impl CompressedCircuit {
+    // Canonical MessagePack uses at most nine bytes per usize, two per u8,
+    // and five for each vector header. For every constraint, the encoder can
+    // add one public-input index, eleven 32-byte scalars, one polynomial with
+    // eleven indices, and one constraint with five indices.
+    const PACKED_FIXED_BYTES: usize = 30;
+    const PACKED_BYTES_PER_CONSTRAINT: usize = 857;
+
+    const SELECTORS_PER_POLYNOMIAL: usize = 11;
```
moCello, line 105: **NOTE:** The packed-size constants restate the msgpacker wire encoding and the compressed-struct shape by hand; nothing ties them to either, so a later change can silently invalidate them.

**Context:** The values are correct for msgpacker 0.4.8 (`[u8;32]` packs headerless at ≤64 bytes, usize ≤9, array headers ≤5): 9 + 11×64 + 11×9 + 5×9 = 857 and 1 + 9 + 4×5 = 30. The version pin `=0.4.8` holds the encoding side. The struct side is looser: adding a selector to `CompressedPolynomial`, a field to `CompressedConstraint`, or reordering the fields `unpack_bounded` lists would leave the constants stale without a compile error. `SELECTORS_PER_POLYNOMIAL` and the `[usize; 11]` return type of `scalar_indices` are the same fact stated twice. The bound also has no slack to absorb drift: once all four collections pass 65535 elements and every array header reaches its 5-byte form, the limit is exactly saturated.

**Impact:** An undersized limit rejects legitimate canonical circuits (a compatibility break at the decode boundary); an oversized one only loosens the allocation bound.

**Fix:** Derive the constants from named parts (max packed usize, packed u8, header size, `SELECTORS_PER_POLYNOMIAL`, field counts) as a const expression, size `scalar_indices`'s return array with `SELECTORS_PER_POLYNOMIAL`, and add a regression test that packs a worst-case constraint (11 distinct scalars with high bytes, large indices) and asserts the encoding fits inside `PACKED_BYTES_PER_CONSTRAINT` plus `PACKED_FIXED_BYTES`.

src/compiler.rs
```diff
@@ -80,16 +82,34 @@ impl Compiler {
...
+            .max_degree()
+            .saturating_sub(PublicParameters::ADDED_BLINDING_DEGREE);
+        let max_domain_size = if available == 0 {
+            0
+        } else {
+            1usize << (usize::BITS - available.leading_zeros() - 1)
+        };
+
+        max_domain_size.saturating_sub(Self::CIRCUIT_SIZE_PADDING)
+    }
```
moCello, line 111: **NOTE:** This is a hand-derived inverse of the capacity rule `compile_with_composer` + `trim`/`truncate` enforce, and no test pins the two to each other at the boundary.

**Context:** The derivation is currently exactly tight: compilation succeeds iff `(constraints + 6).next_power_of_two() + ADDED_BLINDING_DEGREE <= max_degree`, and this function returns the largest constraint count satisfying that. If `trim`, `CommitKey::truncate`, or the domain-size computation ever changes, this bound drifts silently — too strict rejects compilable compressed circuits, too loose re-opens allocation headroom. `capacity_limits_are_inclusive` tests the decoder bound against a hard-coded 2, not against a bound derived from real public parameters, and `compiler_accepts_sparse_witness_labels` compiles 1 constraint against a 2-capacity setup — neither exact capacity nor capacity-plus-one goes through `compile_with_compressed`.

**Fix:** Add a boundary test through the public path: for a small `PublicParameters`, a compressed circuit with exactly `max_constraints(pp)` constraints compiles via `compile_with_compressed`, and one more constraint is rejected.

src/composer/compress.rs
```diff
@@ -397,11 +455,83 @@ impl CompressedCircuit {
...
+            0xdd => {
+                let bytes = self.take(4)?;
+                usize::try_from(u32::from_be_bytes([
+                    bytes[0], bytes[1], bytes[2], bytes[3],
+                ]))
+                .map_err(|_| Error::InvalidCompressedCircuit)
+            }
+            _ => Err(Error::InvalidCompressedCircuit),
+        }
+    }
```
moCello, line 513: **NOTE:** The rejection arms of the new header parser are untested — no test feeds a non-array tag, a truncated header, or a 0xdd (array32) length.

**Context:** The fixtures exercise fixarray, and the 23-scalar case reaches the `0xdc` branch — though it is then rejected by `unpack_vec`'s `len > max_len` check, not by this function's own arms. No test reaches the `_ => Err(...)` arm, the `take` short-buffer path, or the `0xdd` branch (live for real circuits above 65535 elements). These are the paths a corrupt or adversarial payload actually hits.

**Fix:** Add inline unit tests driving `PackedCircuitReader` with raw bytes: a bin/map/int tag where an array is expected, a 0xdc header cut short, and a 0xdd header with an in-bounds length.

moCello (approved):
APPROVE — The capacity derivation is exactly tight against `trim`, the packed-size constants are correct upper bounds for the pinned msgpacker encoding, the witness remap provably cannot change key bytes (sigma permutations are position-indexed), and the byte-identity test locks the compatibility claim; nothing gates.

Three notes below, all yours to take or leave: two harden hand-mirrored invariants (packed-size constants, capacity inverse) with derivations or boundary tests, one adds tests for the new parser's rejection arms.

The CHANGELOG link-definition block now reads #935, #931, #933 — one line out of the descending order the block otherwise keeps.

PR #955, opened 2026-08-27
Handle fixed-buffer serialization errors

### LINKED ISSUE #959: Size serialized ProverKey buffers from each polynomial length
## Summary

ProverKey::serialization_size assumes every serialized polynomial has the same coefficient count as q_m. Selector polynomials are canonicalized independently, so their actual lengths can differ.

This can over-allocate the prover-key buffer and emit trailing zero padding when q_m is longer, or under-allocate and truncate or fail when another polynomial is longer.

## Expected behavior

Size the buffer from each serialized polynomial actual length so valid uneven proving keys serialize without padding or truncation and round-trip byte-for-byte.

Neotamandua:
### Summary

Handle fixed-buffer serialization failures explicitly instead of silently ignoring `dusk_bytes::Write` errors.

-  adds a crate-local `BufferWriter` that fails fast if a serialization buffer is undersized
- removes `unused_must_use` suppressions from fixed-size serializers
- fixes `ProverKey` serialization sizing to account for actual polynomial lengths
- adds regression coverage for uneven polynomial lengths

Public API shape is unchanged. Valid serialization behavior is preserved, while previously silent buffer-size mismatches now fail as invariant failures

Resolves #959

### DIFF
--- a/src/commitment_scheme/kzg10/key.rs
+++ b/src/commitment_scheme/kzg10/key.rs
@@ -29,7 +29,7 @@ use super::proof::Proof;
 use crate::error::Error;
 use crate::fft::Polynomial;
 use crate::transcript::TranscriptProtocol;
-use crate::util;
+use crate::{BufferWriter, util};
 
 /// CommitKey is used to commit to a polynomial which is bounded by the
 /// max_degree.
@@ -593,15 +593,13 @@ fn batch_challenge(
 impl Serializable<{ G1Affine::SIZE + G2Affine::SIZE * 2 }> for OpeningKey {
     type Error = dusk_bytes::Error;
 
-    #[allow(unused_must_use)]
     fn to_bytes(&self) -> [u8; Self::SIZE] {
-        use dusk_bytes::Write;
         let mut buf = [0u8; Self::SIZE];
-        let mut writer = &mut buf[..];
-        // This can't fail therefore we don't care about the Result nor use it.
+        let mut writer = BufferWriter::new(&mut buf);
         writer.write(&self.g.to_bytes());
         writer.write(&self.h.to_bytes());
         writer.write(&self.x_h.to_bytes());
+        writer.finish();
 
         buf
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,6 +50,8 @@
 #![deny(missing_docs)]
 #![cfg_attr(not(feature = "std"), no_std)]
 
+pub(crate) use buffer_writer::BufferWriter;
+
 cfg_if::cfg_if!(
 if #[cfg(feature = "alloc")] {
     /// `macro_use` will declare `vec!`. However, if `libstd` is present, then this
@@ -72,6 +74,7 @@ if #[cfg(feature = "alloc")] {
 #[cfg(feature = "debug")]
 pub(crate) mod debugger;
 
+mod buffer_writer;
 mod commitment_scheme;
 mod error;
 mod fft;
--- a/src/proof_system/widget.rs
+++ b/src/proof_system/widget.rs
@@ -6,6 +6,7 @@
 
 use dusk_bytes::{DeserializableSlice, Serializable};
 
+use crate::BufferWriter;
 use crate::commitment_scheme::Commitment;
 
 pub mod arithmetic;
@@ -84,11 +85,9 @@ impl<C> CheckBytes<C> for ArchivedVerifierKey {
 impl Serializable<{ 20 * Commitment::SIZE + u64::SIZE }> for VerifierKey {
     type Error = dusk_bytes::Error;
 
-    #[allow(unused_must_use)]
     fn to_bytes(&self) -> [u8; Self::SIZE] {
-        use dusk_bytes::Write;
         let mut buff = [0u8; Self::SIZE];
-        let mut writer = &mut buff[..];
+        let mut writer = BufferWriter::new(&mut buff);
 
         writer.write(&(self.n as u64).to_bytes());
         writer.write(&self.arithmetic.q_m.to_bytes());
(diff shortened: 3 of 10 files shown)


HDauven (approved):
LGTM

src/proof_system/widget.rs
```diff
@@ -339,20 +356,20 @@ pub(crate) mod alloc {
...
+
             let size = self.serialization_size();
             let eval_size = self.arithmetic.q_m.1.evals.len() * BlsScalar::SIZE
                 + EvaluationDomain::SIZE;
 
             let mut bytes = vec![0u8; size];
 
-            let mut writer = &mut bytes[..];
+            let mut writer = BufferWriter::new(&mut bytes);
             writer.write(&(self.n as u64).to_bytes());
```
moCello, line 373: **CAUTION:** `ProverKey::to_var_bytes` never calls `writer.finish()`, so an over-estimated `serialization_size` still passes silently

**Context:** Every fixed-size `to_bytes` in this PR closes with `finish()`; this is the one writer whose buffer size is computed at runtime, so it is the one where the check does the most work. `serialization_size` derives `eval_size` from `q_m` alone and assumes every evaluation set shares that length, while `to_var_bytes` writes each set's own `to_var_bytes()`. Today every set is built over the 8n coset domain, so they agree; if that ever drifts, an under-estimate panics but an over-estimate leaves trailing zero bytes that `ProverKey::from_slice` does not reject — the same shape as the pre-fix padding. `eval_size` is also computed twice (here and in `serialization_size`), which is a second way for the two to disagree.

**Fix:** Add `writer.finish();` before `bytes` at the end of `to_var_bytes`, and move the `eval_size` computation into one private helper on `ProverKey` that both `serialization_size` and `to_var_bytes` call.

src/proof_system/widget.rs
```diff
@@ -319,16 +322,30 @@ pub(crate) mod alloc {
         /// Duplicate polynomials of the ProverKey (e.g. `q_L`, `q_R` and `q_C`)
         /// are only counted once.
         fn serialization_size(&self) -> usize {
-            // Fetch size in bytes of each Polynomial
-            let poly_size = self.arithmetic.q_m.0.len() * BlsScalar::SIZE;
+            let polynomial_lengths = [
```
moCello, line 329: **CAUTION:** No `CHANGELOG.md` entry and no tracking issue for the prover-key sizing fix

**Context:** Unequal coefficient counts are the normal shape of a compiled key — `Polynomial::from_coefficients_vec` trims leading zeros from every selector. Before this change `Prover::to_bytes` therefore padded the prover-key segment with zeros whenever `q_m` was the longest selector, and truncated it whenever another selector was longer. Both are consumer-visible: `serialized_size` and the emitted bytes change for the padded case (old bytes still decode), and previously undecodable provers now roundtrip. Nearly every entry in `[Unreleased]` carries an issue link, so an unlinked one stands out.

**Fix:** Open a tracking issue for the sizing bug, add under `### Fixed`: `- Size serialized ProverKey buffers from each polynomial's actual length [#NNN]` and the matching link definition at the bottom, and link the PR to the issue.

src/lib.rs
```diff
@@ -50,6 +50,51 @@
...
+            .expect("serialization buffer is too small");
+    }
+
+    pub(crate) fn finish(self) {
+        assert!(
+            self.buffer.is_empty(),
+            "serialization buffer was not fully written"
+        );
+    }
+}
```
moCello: **NOTE:** `BufferWriter` and its test module live in the crate root instead of their own module

**Context:** `src/lib.rs` is otherwise attributes, module declarations and re-exports. `util.rs` is the natural home but is `alloc`-gated, and `BufferWriter` is needed by non-`alloc` serializers (`OpeningKey`, `EvaluationDomain`). Also `use dusk_bytes::Write;` sits inside `write()` rather than at module top.

**Fix:** Move the type and its tests to a new `src/buffer_writer.rs` (`mod buffer_writer;` at the root, no feature gate) with `use dusk_bytes::Write;` at the top of that file.

src/commitment_scheme/kzg10/key.rs
```diff
@@ -593,15 +593,15 @@ fn batch_challenge(
 impl Serializable<{ G1Affine::SIZE + G2Affine::SIZE * 2 }> for OpeningKey {
     type Error = dusk_bytes::Error;
 
-    #[allow(unused_must_use)]
     fn to_bytes(&self) -> [u8; Self::SIZE] {
-        use dusk_bytes::Write;
+        use crate::BufferWriter;
```
moCello: **NOTE:** `use crate::BufferWriter;` placed inside function bodies at seven sites

**Context:** Same pattern in `fft/domain.rs`, `proof_system/proof.rs`, `proof_system/linearization_poly.rs`, `proof_system/widget.rs` (two sites) and `proof_system/widget/arithmetic/verifierkey.rs`. The replaced `use dusk_bytes::Write;` had the same shape, so this is inherited, but the new lines are the ones being added.

**Fix:** Hoist each `use crate::BufferWriter;` to the module's import block.

moCello (commented):
COMMENT — The fixed-size serializers and the prover-key sizing fix are correct and the regression test fails on the old code; two cautions ask for a change here and are worth one push together.
- `ProverKey::to_var_bytes` is the one writer with a runtime-computed size and the one missing `finish()`.
- The sizing fix is consumer-visible and needs a `CHANGELOG.md` entry backed by a tracking issue.
- Two notes on placement of `BufferWriter` and in-body imports; take them in the same push or leave them.

moCello (approved):
APPROVE — every ask from the prior review is implemented: `to_var_bytes` now closes with `finish()` and shares one `evaluations_serialization_size` helper, the sizing fix is recorded in the CHANGELOG against #959, `BufferWriter` has its own module, and all imports sit at module top; the new tests cover both the uneven-length roundtrip and the short-evaluations panic path, and nothing new asks for a change.

For the record, not a change request: `serialization_size` counts `coeffs.len()` while `Polynomial::to_var_bytes` writes `degree + 1` coefficients — they agree because every constructor and decode path trims trailing zeros, and any future drift now panics at `finish()` rather than corrupting output, so the invariant is runtime-enforced by this PR.

PR #956, opened 2026-08-30
fix: bind prover public inputs to compiled rows

HDauven:
## Summary

Persist the compiled public-input row layout in `Prover` and use that layout while constructing proofs.

The serialized prover format receives an explicit magic value and row-index section so incompatible legacy prover blobs are rejected instead of being interpreted with the wrong layout.

## Why

Reconstructing public-input rows from the proving-time composer can bind values to different rows from those used during compilation. The compiled layout is the authoritative mapping.

## Performance

An exact-head warm Criterion A/B run on the same ARM64 Mac measured a 2^10-gate proof at:

- base: 175.99 ms
- patch: 175.96 ms
- delta: -0.02%

Criterion reported no statistically significant change (p = 0.98). The proving path no longer allocates and sorts a temporary index vector merely to compare public-input counts, but cryptographic work dominates the benchmark.

Each prover gains a 24-byte vector header plus 8 bytes per public-input row. Serialized provers grow by 16 + 8 bytes per public-input row.

## Validation

- `cargo fmt --check`
- six focused public-input tests
- both `assert_scalar` integration tests

Draft pending serialized-format and compatibility review.

Related investigation: https://github.com/HDauven/backlog/issues/118

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Bind proving-time public inputs to the rows recorded at circuit compilation.
+  Serialized provers must be regenerated because the format now includes that
+  row layout.
 - Normalize unrepresentable serialized prover-key lengths as invalid data
   [#937]
 - Bound compressed-circuit decompression and reconstruction allocation by the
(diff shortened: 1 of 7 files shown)


HDauven:
Superseded by #958, which uses a repository-appropriate branch name and links the corresponding repository issue and changelog entry.

Issue #940, opened 2026-08-20
Normalize serialized-length conversions and truncation errors


HDauven:
## Summary

Remaining serialized-length conversion and truncation-error consistency work from the review of #938.

At master `6fecc3a72aa190efed1c169c325f75bc1f821e61`, `VerifierKey::from_bytes` and `Verifier::try_from_bytes` still convert serialized `u64` values with truncating `as usize` casts, which can map distinct serialized values to the same value on 32-bit targets. The corresponding outer prover length and public-input-index conversions are already checked.

Truncation is reported through direct `Error::NotEnoughBytes` and wrapped `Error::BytesError(dusk_bytes::Error::BadLength { found, expected })`. The original review and earlier issue text incorrectly named the nested variant `dusk_bytes::Error::NotEnoughBytes`.

## Error-policy decision

Preserve the existing distinction instead of discarding the nested reader's `found`/`expected` diagnostics. Document it and test truncated headers and payloads. Values that cannot be represented on the target return `Error::BytesError(dusk_bytes::Error::InvalidData)`, consistent with the checked prover conversions. Verifier byte-length multiplication and addition overflow now also return `Error::BytesError(dusk_bytes::Error::InvalidData)`, matching the outer prover decoder. This deliberately changes the previous `Error::NotEnoughBytes` outcome for those malformed headers, including on 64-bit targets. Representable lengths with missing bytes retain the existing short-read errors. Other decoders' arithmetic-error policies are outside this expansion.

Downstream adoption note: callers and regression tests matching `NotEnoughBytes` for verifier arithmetic overflow must update their expected variant, including Rusk's decoder-error regression. The overflow inputs remain rejected; this follow-up changes their diagnostic, not the verifier's acceptance conditions. No downstream dependency rollout or historical replay validation is included.

## Already addressed

- #958 checks outer prover length/index conversions and binds proof construction to the compiled public-input count and rows.
- #961 validates verifier size metadata, rejects out-of-range, duplicate or descending public-input indexes, and rejects trailing serialized data. This completes the verifier-layout scope added in the earlier issue comment; do not repeat that implementation here.

## Remaining acceptance criteria

- Replace remaining truncating serialized-length and public-input-index conversions, including those in `VerifierKey::from_bytes` and `Verifier::try_from_bytes`, with checked conversions consistent with the existing prover policy.
- Reject values that cannot be represented on the target as malformed data without unwinding.
[...]

PR #958, opened 2026-08-30
Bind proofs to compiled public-input rows

### LINKED ISSUE #957: Bind prover public inputs to compiled circuit rows
## Context

The proving path reconstructs public-input row positions from the proving-time composer. A circuit instance can preserve the public-input count while moving values to different rows than the layout used when the prover was compiled.

## Objective

Persist the compiled public-input row layout in `Prover` and use it as the authoritative mapping during proof construction.

## Acceptance criteria

- [ ] Proving rejects public inputs moved away from compiled rows.
- [ ] Proving rejects a different public-input count.
- [ ] Serialized provers persist and validate the compiled row layout.
- [ ] Legacy/incompatible serialized provers fail explicitly.
- [ ] Deserialization validates ordering, bounds, lengths, and trailing data.
- [ ] Existing correctly structured circuits continue proving and verifying.
- [ ] No measurable proving-time regression is introduced.

HDauven:
## Summary

Persist the compiled public-input row layout in `Prover` and use that layout while constructing proofs.

The serialized prover format receives an explicit magic value and row-index section so incompatible legacy prover blobs are rejected instead of being interpreted with the wrong layout.

## Why

Reconstructing public-input rows from the proving-time composer can bind values to different rows from those used during compilation. The compiled layout is the authoritative mapping.

## Performance

An exact-head warm Criterion A/B run on the same ARM64 Mac measured a 2^10-gate proof at:

- base: 175.99 ms
- patch: 175.96 ms
- delta: -0.02%

Criterion reported no statistically significant change (p = 0.98). The proving path no longer allocates and sorts a temporary index vector merely to compare public-input counts, but cryptographic work dominates the benchmark.

On 64-bit targets, each prover gains a 24-byte vector header plus 8 bytes per public-input row. Serialized provers grow by 16 + 8 bytes per public-input row.

## Validation

- `cargo fmt --all --check`;
- focused public-input layout tests, including truncated versus trailing row-index sections, overflowing aggregate lengths, duplicate and descending serialized indexes, positive prove/verify after decoding, and rejection when runtime public inputs move away from compiled rows;
- both `assert_scalar` integration tests;
- complete GitHub CI is green on the current head, including tests, Clippy, docs, examples, benchmarks, Cargo Deny, `no_std`, and the 32-bit target.

Closes #957

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -788,7 +788,11 @@ is necessary since `rkyv/validation` was required as a bound.
 - Proof system module.
 
 <!-- ISSUES -->
+<<<<<<< HEAD
 [#959]: https://github.com/dusk-network/plonk/issues/959
+=======
+[#957]: https://github.com/dusk-network/plonk/issues/957
+>>>>>>> 7fdedc29 (Bind proofs to compiled public-input rows)
 [#953]: https://github.com/dusk-network/plonk/issues/953
 [#951]: https://github.com/dusk-network/plonk/issues/951
 [#950]: https://github.com/dusk-network/plonk/issues/950
--- a/src/compiler/prover.rs
+++ b/src/compiler/prover.rs
@@ -39,7 +39,13 @@ pub struct Prover {
     vanishing_coset_inverses: [BlsScalar; 8],
     pub(crate) size: usize,
     pub(crate) constraints: usize,
+    public_input_indexes: Vec<usize>,
 }
+
+// The low 32 bits are deliberately all ones. Previous 32-bit decoders read
+// this as the label length and must reject new-format provers instead of
+// truncating away a version bit and silently dropping the row layout.
+const PROVER_FORMAT_V2_MAGIC: u64 = 0x5052_5632_FFFF_FFFF;
 
 impl ops::Deref for Prover {
     type Target = ProverKey;
@@ -57,9 +63,16 @@ impl Prover {
         verifier_key: VerifierKey,
         size: usize,
         constraints: usize,
+        public_input_indexes: Vec<usize>,
     ) -> Result<Self, Error> {
         if constraints.checked_next_power_of_two() != Some(size)
             || prover_key.n != size
+            || public_input_indexes
+                .iter()
+                .any(|index| *index >= constraints)
+            || public_input_indexes
+                .windows(2)
+                .any(|indexes| indexes[0] >= indexes[1])
         {
             return Err(dusk_bytes::Error::InvalidData.into());
         }
@@ -111,6 +124,7 @@ impl Prover {
             vanishing_coset_inverses,
             size,
             constraints,
+            public_input_indexes,
         })
     }
 
@@ -223,8 +237,14 @@ impl Prover {
         let commit_key_len = commit_key.len();
         let verifier_key_len = verifier_key.len();
 
-        let size =
-            48 + label_len + prover_key_len + commit_key_len + verifier_key_len;
+        let public_input_indexes_len =
+            self.public_input_indexes.len() * u64::SIZE;
+        let size = 8 * u64::SIZE
+            + label_len
+            + prover_key_len
+            + commit_key_len
+            + verifier_key_len
+            + public_input_indexes_len;
 
         (size, prover_key, commit_key, verifier_key)
     }
@@ -247,6 +267,7 @@ impl Prover {
         let size = self.size as u64;
         let constraints = self.constraints as u64;
 
+        bytes.extend(PROVER_FORMAT_V2_MAGIC.to_be_bytes());
         bytes.extend(label_len.to_be_bytes());
         bytes.extend(prover_key_len.to_be_bytes());
         bytes.extend(commit_key_len.to_be_bytes());
@@ -258,6 +279,10 @@ impl Prover {
         bytes.extend(prover_key);
         bytes.extend(commit_key);
         bytes.extend(verifier_key);
+        bytes.extend((self.public_input_indexes.len() as u64).to_be_bytes());
+        for index in &self.public_input_indexes {
+            bytes.extend((*index as u64).to_be_bytes());
+        }
 
         bytes
     }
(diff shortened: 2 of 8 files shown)


tests/common/mod.rs
```diff
@@ -48,13 +46,13 @@ pub(crate) fn check_satisfied_circuit_fails<C, R>(
...
-        .expect("Prover for valid circuit shouldn't fail");
-
-    assert_eq!(*pi_expected, pi_circuit);
-
-    verifier.verify(&proof, &pi_expected).expect_err(msg);
+        .expect_err("proving should reject a mismatched public-input count");
+    assert!(
+        matches!(error, Error::InconsistentPublicInputsLen { .. }),
+        "{msg}: got {error:?} for public inputs {pi_expected:?}"
+    );
```
moCello, line 57: **CAUTION:** The verifier's `InconsistentPublicInputsLen` path loses its only test coverage.

**Context:** The four rewritten `assert_scalar` call sites were the only tests reaching the count check in `Verifier::verify_with_version`; they now stop at prove-time rejection. No remaining test calls `verify` with a public-input vector whose length differs from the compiled layout.

**Fix:** Add one direct assertion that `verifier.verify(&proof, &wrong_len_pi)` returns `InconsistentPublicInputsLen` — either in this helper (prove with the compiled-shape circuit, verify with a truncated/extended vector) or as a unit test next to `Verifier::verify_with_version`.

src/compiler/prover.rs
```diff
@@ -326,6 +366,40 @@ impl Prover {
...
+        let public_input_indexes_len =
+            u64::from_be_bytes(bytes[..u64::SIZE].try_into().expect("checked"));
+        let public_input_indexes_len =
+            usize::try_from(public_input_indexes_len)
+                .map_err(|_| dusk_bytes::Error::InvalidData)?;
+        bytes = &bytes[u64::SIZE..];
+
+        if public_input_indexes_len > constraints {
+            return Err(dusk_bytes::Error::InvalidData.into());
+        }
```
moCello, line 385: **NOTE:** New decoder branch has no test: no serialized blob with an index count exceeding `constraints` is exercised.

**Context:** The `malformed_indexes` cases in `prover_serialization_requires_valid_public_input_layout` use counts of 2, below the test circuit's constraint count, so they are rejected by the ordering check in `Prover::new` instead.

**Fix:** Add a `malformed_indexes`-style case whose declared count word exceeds `constraints` — the check fires before the byte-length comparisons, so no index words need to be appended.

CHANGELOG.md
```diff
@@ -92,6 +92,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Bind proving-time public inputs to the rows recorded at circuit compilation.
+  Serialized provers must be regenerated because the format now includes that
+  row layout [#957](https://github.com/dusk-network/plonk/issues/957).
```
moCello: **NOTE:** Inline issue link deviates from the changelog's reference-link convention.

**Context:** Every other entry uses `[#NNN]` with a definition in the link block at the bottom of the file (e.g. `[#937]` on the adjacent line).

**Fix:** Change to `[#957]` and add `[#957]: https://github.com/dusk-network/plonk/issues/957` to the link-definition block.

src/compiler/prover.rs
```diff
@@ -270,43 +295,58 @@ impl Prover {
     {
         let mut bytes = bytes.as_ref();
 
-        if bytes.len() < 48 {
+        if bytes.len() < 56 {
             return Err(Error::NotEnoughBytes);
         }
 
+        let magic = <[u8; 8]>::try_from(&bytes[..8]).expect("checked len");
```
moCello: **NOTE:** New `expect` calls in the library decode path (here and the row-count word read); `split_first_chunk` removes them without cost.

**Context:** Both are guarded by explicit length checks and cannot fire, and they follow the pattern of the pre-existing header parsing. `bytes.split_first_chunk::<8>().ok_or(Error::NotEnoughBytes)?` yields the array and the remainder in one panic-free step, replacing the manual slice-check/`expect`/re-slice triple.

**Fix:** Use `split_first_chunk` for the magic and count reads; yours to take or leave, and applying it to the pre-existing header reads is a separate cleanup.

tests/common/mod.rs
```diff
@@ -31,15 +31,13 @@ pub(crate) fn check_satisfied_circuit<C, R>(
...
+// Check that proving rejects a satisfied circuit whose public-input count does
+// not match the compiled circuit.
 // As this is a very specific test case, this function will not be used by all
 // tests.
 #[allow(dead_code)]
-pub(crate) fn check_satisfied_circuit_fails<C, R>(
+pub(crate) fn check_public_input_count_mismatch<C, R>(
     prover: &Prover,
-    verifier: &Verifier,
     pi_expected: &Vec<BlsScalar>,
```
moCello: **NOTE:** `pi_expected` is only interpolated into the assert message; the helper never checks it against anything.

**Context:** Call sites read as if the prover is checked against `pi_expected`, but only the count mismatch is asserted. The four `assert_scalar` messages still say "should not verify" although the helper now asserts prove-time rejection.

**Fix:** Drop the parameter (or narrow it to `expected_len: usize`), and update the call-site `msg` strings to say proving is rejected.

moCello (commented):
COMMENT — The fix is sound: the compiled layout is now the single authority for public-input rows, the count-plus-lookup check correctly implies the runtime rows equal the compiled ones, the magic word's all-ones low half provably makes legacy 32-bit decoders reject the new format, and decoder allocation stays bounded by actual input length.

- One caution asks for a change here: rewriting the four `assert_scalar` mismatch tests removed the only coverage of the verifier's own public-input count check — one direct `verify`-with-wrong-length assertion restores it.
- Five notes ride along, all small; the changelog link-style fix is a two-line edit worth batching into the same push, and one note asks only for a follow-up issue (the verifier's decode path still has the truncating casts and unvalidated index layout this PR fixed for the prover).

src/compiler/prover.rs
```diff
@@ -270,43 +295,58 @@ impl Prover {
...
+        if u64::from_be_bytes(magic) != PROVER_FORMAT_V2_MAGIC {
+            return Err(dusk_bytes::Error::InvalidData.into());
+        }
+        bytes = &bytes[8..];
+
         let label_len = <[u8; 8]>::try_from(&bytes[..8]).expect("checked len");
-        let label_len = u64::from_be_bytes(label_len) as usize;
+        let label_len = usize::try_from(u64::from_be_bytes(label_len))
+            .map_err(|_| dusk_bytes::Error::InvalidData)?;
```
moCello, line 312: **NOTE:** The hardening applied here — checked `usize::try_from` casts and layout validation — is absent from the verifier's twin decode path; follow-up work, nothing to change in this PR.

**Context:** `Verifier::try_from_bytes` still uses truncating `u64 ... as usize` casts on every header word, and `Verifier::new` accepts `public_input_indexes` without the bounds or strict-ascending checks `Prover::new` now enforces, and without a trailing-data check. Verifier blobs travel the same storage channels as prover blobs, so a hand-edited or truncated-on-32-bit verifier blob is decoded with weaker guarantees than a prover blob.

**Fix:** Track in a follow-up issue: mirror the checked casts and layout validation in `Verifier::try_from_bytes` / `Verifier::new`; nothing to change in this PR.

src/compiler/verifier.rs
```diff
@@ -330,4 +330,26 @@ mod tests {
...
+
+        assert!(public_inputs.is_empty());
+        assert_eq!(
+            verifier.verify(&proof, &[BlsScalar::from(1u64)]),
+            Err(Error::InconsistentPublicInputsLen {
+                expected: 0,
+                provided: 1,
+            })
+        );
+    }
```
moCello, line 354: **NOTE:** Verifier-side layout validation is still untracked — the cast cleanup is already covered by #940, the rest needs an issue; nothing to change in this PR.

**Context:** This test restores the verify-path count coverage, and [#940](https://github.com/dusk-network/plonk/issues/940) already tracks the truncating `u64 as usize` casts in `Verifier::try_from_bytes`. Not tracked anywhere: `Verifier::new` accepts `public_input_indexes` without the bounds, strict-ascending, or count-vs-constraints checks `Prover::new` now enforces, and the verifier decoder accepts trailing bytes. Verifier blobs travel the same storage channels as prover blobs.

**Fix:** Extend #940, or file a sibling issue, to mirror the layout validation now in `Prover::new` / `Prover::try_from_bytes` on the verifier path.

moCello (approved):
APPROVE — Every prior finding that asked for a change on this branch is fixed and verified; the one remaining item asks only for follow-up issue tracking, not a change here.

- The verifier count-mismatch test, the excessive-count decoder case, the changelog link, the `split_first_chunk` reads, and the test-helper rework all landed as requested.
- Still open: verifier-side layout validation (`Verifier::new` bounds and ordering checks, trailing-data rejection) has no tracking issue; the truncating-cast half is already covered by #940. The inline note carries the details — extending #940 or filing a sibling issue closes out this review.

Issue #913, opened 2026-08-10
Reject trailing data in compressed circuits


HDauven:
## Summary

`CompressedCircuit::from_bytes` inflates the supplied bytes and decodes one MessagePack circuit value, but does not verify that the decoder consumed the entire inflated payload. A valid compressed circuit can therefore be extended with arbitrary trailing bytes, recompressed, and accepted by `Compiler::compile_with_compressed`.

## Impact and current behavior

Malformed compressed-circuit artifacts are accepted instead of returning `Error::InvalidCompressedCircuit`. The trailing bytes do not affect the circuit currently compiled, but accepting non-canonical payloads weakens artifact-boundary validation and permits unnoticed or ambiguous extra data to travel with a circuit.

## Acceptance criteria

- Compressed-circuit decoding requires the MessagePack circuit value to consume the entire inflated payload.
- A valid circuit followed by any extra MessagePack value is rejected with `Error::InvalidCompressedCircuit`.
- Valid compressed circuits continue to compile, prove, and verify normally.
- A regression test covers trailing-data rejection and the valid-circuit control.

PR #961, opened 2026-08-31
fix(plonk): Validate verifier public-input layouts

### LINKED ISSUE #960: Harden verifier public-input layout deserialization
## Summary

`Prover::new` and `Prover::try_from_bytes` validate the compiled public-input
row layout, but the equivalent verifier path does not enforce the same
invariants. `Verifier::try_from_bytes` also accepts undeclared trailing bytes.

As a result, malformed or corrupted verifier blobs can construct verifier
state with duplicate, descending, or out-of-range public-input rows, and
multiple byte encodings can decode to the same verifier.

The truncating `u64 as usize` conversions in this decoder are already tracked
by #940. This issue covers the remaining verifier-layout and exact-consumption
work identified during review of #958.

## Acceptance criteria

- Validate in `Verifier::new` that:
  - `constraints.checked_next_power_of_two() == Some(size)`;
  - `verifier_key.n == constraints`;
  - the public-input count does not exceed `constraints`;
  - every public-input index is below `constraints`;
  - indexes are strictly increasing, rejecting duplicates and descending
    layouts.
- Make `Verifier::try_from_bytes` reject undeclared trailing bytes.
- Preserve valid verifier serialization and proof verification behavior.
- Add regression tests for excessive counts, duplicate, descending, and
  out-of-range indexes, incoherent size metadata, trailing bytes, and a valid
  round trip with public inputs.
- Preserve `no_std` support.

Follow-up to #958 and complementary to #940.

HDauven:
## Summary

- validate verifier size metadata and public-input row layouts during
  construction;
- reject excessive, duplicate, descending, and out-of-range public-input
  indexes;
- reject undeclared trailing bytes during verifier deserialization;
- cover malformed layouts and a positive serialized-verifier prove/verify
  round trip.

## Stack

This PR is stacked on #958 and should be merged after it. Keeping this work
separate leaves the already-reviewed public-input binding fix focused and ready
to merge.

## Performance

Validation is linear in the number of public inputs and runs only when a
verifier is constructed or deserialized. Proof generation and verification hot
paths are unchanged.

## Validation

- `cargo fmt --all --check`
- `cargo test --release` on macOS ARM: 147 unit tests and all integration tests
  passed

Closes #960.

### DIFF
--- a/src/compiler/verifier.rs
+++ b/src/compiler/verifier.rs
@@ -42,7 +42,6 @@ impl Verifier {
 
         if constraints.checked_next_power_of_two() != Some(size)
             || verifier_key.n != constraints
-            || public_input_indexes.len() > constraints
             || public_input_indexes
                 .iter()
                 .any(|index| *index >= constraints)
--- a/src/compiler/verifier.rs
+++ b/src/compiler/verifier.rs
@@ -39,6 +39,19 @@ impl Verifier {
         constraints: usize,
     ) -> Result<Self, Error> {
         let domain = EvaluationDomain::new(verifier_key.n)?;
+
+        if constraints.checked_next_power_of_two() != Some(size)
+            || verifier_key.n != constraints
+            || public_input_indexes.len() > constraints
+            || public_input_indexes
+                .iter()
+                .any(|index| *index >= constraints)
+            || public_input_indexes
+                .windows(2)
+                .any(|indexes| indexes[0] >= indexes[1])
+        {
+            return Err(dusk_bytes::Error::InvalidData.into());
+        }
         let public_input_roots = public_input_indexes
             .iter()
             .map(|index| domain.group_gen_inv.pow(&[*index as u64, 0, 0, 0]))
@@ -160,6 +173,10 @@ impl Verifier {
         let constraints = u64::from_be_bytes(constraints) as usize;
         bytes = &bytes[8..];
 
+        if public_input_indexes_len > constraints {
+            return Err(dusk_bytes::Error::InvalidData.into());
+        }
+
         let required_len = label_len
             .checked_add(verifier_key_len)
             .and_then(|len| len.checked_add(opening_key_len))
@@ -168,6 +185,9 @@ impl Verifier {
 
         if bytes.len() < required_len {
             return Err(Error::NotEnoughBytes);
+        }
+        if bytes.len() > required_len {
+            return Err(dusk_bytes::Error::InvalidData.into());
         }
 
         let label = &bytes[..label_len];
@@ -287,6 +307,36 @@ mod tests {
         }
     }
 
+    #[derive(Default)]
+    struct PublicInputCircuit;
+
+    impl Circuit for PublicInputCircuit {
+        fn circuit(&self, composer: &mut Composer) -> Result<(), Error> {
+            for value in [7u64, 11] {
+                let value = BlsScalar::from(value);
+                let witness = composer.append_witness(value);
+                composer.assert_equal_constant(witness, 0u64, Some(value));
+            }
+            Ok(())
+        }
+    }
+
+    fn header_word(bytes: &[u8], index: usize) -> usize {
+        let offset = index * u64::SIZE;
+        u64::from_be_bytes(
+            bytes[offset..offset + u64::SIZE]
+                .try_into()
+                .expect("header is complete"),
+        ) as usize
+    }
+
+    fn assert_invalid_verifier(bytes: &[u8]) {
+        assert!(matches!(
+            Verifier::try_from_bytes(bytes),
+            Err(Error::BytesError(dusk_bytes::Error::InvalidData))
+        ));
+    }
+
     #[test]
     fn verifier_try_from_bytes_rejects_overflow_lengths_without_panicking() {
         let mut bytes = Vec::with_capacity(48);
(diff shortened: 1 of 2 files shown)


src/compiler/verifier.rs
```diff
@@ -169,6 +186,9 @@ impl Verifier {
         if bytes.len() < required_len {
             return Err(Error::NotEnoughBytes);
         }
+        if bytes.len() > required_len {
+            return Err(dusk_bytes::Error::InvalidData.into());
+        }
```
moCello, line 190: **CAUTION:** Observable behavior change — `try_from_bytes` now rejects inputs it previously accepted — with no CHANGELOG entry.

**Context:** Trailing bytes and malformed layouts that used to deserialize now fail. That is the point of the change, but it is visible to every consumer of the crate, and the changelog records equivalent hardening under `[Unreleased]` → `Fixed` (e.g. the trailing-data reject for compressed circuits, #913). While editing CHANGELOG.md: it currently carries committed merge-conflict markers around the issue-link block (lines 791–795). They landed on master via #958 and are still unresolved there, so this file needs the markers cleared as well as the new entry.

**Fix:** Add an `[Unreleased]` → `Fixed` entry, e.g. "Validate verifier size metadata and public-input row layouts on construction and deserialization [#960]", with the issue link in the bottom link block.

src/compiler/verifier.rs
```diff
@@ -39,6 +39,19 @@ impl Verifier {
         constraints: usize,
     ) -> Result<Self, Error> {
         let domain = EvaluationDomain::new(verifier_key.n)?;
+
+        if constraints.checked_next_power_of_two() != Some(size)
+            || verifier_key.n != constraints
+            || public_input_indexes.len() > constraints
```
moCello: **NOTE:** The two checks after this one already imply it, so it can never fire.

**Context:** The bounds check rejects any index `>= constraints` and the ordering check rejects `indexes[0] >= indexes[1]`, so any list reaching the end of the disjunction is strictly increasing below `constraints` and therefore has at most `constraints` elements — a longer one trips bounds or ordering first, whatever the caller. Neither caller can deliver an oversized layout anyway: `try_from_bytes` rejects `public_input_indexes_len > constraints` at its early count check before reaching `new`, and the compile path produces sorted unique in-range indexes. This is also the third hand-maintained copy of the layout invariant — `Prover::new` and the compressed-circuit decompression check in `composer/compress.rs` hold the other two — and the copies have already diverged: `Prover::new` omits this condition.

**Fix:** Drop the `public_input_indexes.len() > constraints` condition; the early check in `try_from_bytes` keeps the fast-fail for oversized declared counts. If the layout rules ever change, consider extracting the shared bounds-and-ordering predicate into one helper used by all three sites so they cannot drift further.

moCello (commented):
COMMENT — The validation is correct and complete: every honestly compiled verifier still passes (metadata and index invariants match what `Compiler::preprocess` produces), the seven malformed-blob cases together cover every reachable reject branch, and the round-trip test covers the positive path; what asks for a change is the missing CHANGELOG entry for a behavior change consumers will see.

- One caution: add the `[Unreleased]` changelog entry for the stricter deserialization, linked to #960.
- One note: the `len() > constraints` condition in `Verifier::new` never fires — `try_from_bytes` pre-filters it and the bounds and ordering checks imply it — yours to drop or keep.
- Outside this diff: master carries unresolved merge-conflict markers in CHANGELOG.md (lines 791–795), landed via #958. They are on master now rather than ahead of it, so they want clearing in whichever PR next touches that file — this one, if you add the entry above.

src/compiler/verifier.rs
HDauven: Addressed in 8ce7aaa3. Added the Unreleased/Fixed entry for verifier layout validation and trailing-data rejection, linked #960, and removed the committed conflict markers while retaining both existing issue references.

src/compiler/verifier.rs
HDauven: Addressed in 737a3079. Removed the redundant constructor count condition; the bounds and strict-ordering checks remain, as does the decoder’s early count rejection. Default and all-feature test suites, Clippy, and nightly formatting checks pass locally. Keeping shared-helper extraction out of this change.

src/compiler/verifier.rs
```diff
@@ -169,6 +185,9 @@ impl Verifier {
         if bytes.len() < required_len {
             return Err(Error::NotEnoughBytes);
         }
+        if bytes.len() > required_len {
+            return Err(dusk_bytes::Error::InvalidData.into());
+        }
```
moCello, line 190: **NOTE:** The blob is bounded against the declared lengths, but those lengths are never pinned to `VerifierKey::SIZE` / `OpeningKey::SIZE`, so a verifier still has many encodings.

**Context:** The unpinned length words predate this change; the trailing check is what makes them worth naming. `DeserializableSlice::from_slice` reads the first `SIZE` bytes of a longer slice and ignores the rest, and neither key type overrides it, so raising `verifier_key_len` by N and padding the key segment by N keeps `required_len` self-consistent and decodes to the same verifier. A second channel needs no length change at all: `VerifierKey::to_bytes` writes `n`, fifteen commitments, then `5 * Commitment::SIZE` zero bytes for the retired lookup slots, while `from_bytes` reads back only `n` and the fifteen commitments — 240 of its 968 bytes are never read. The prover decoder's `verifier_key_len` has the same shape. There is no soundness effect: the mapping is many-to-one, so a byte comparison against a known-good blob fails closed, and rusk's `host_verify_plonk` keys its result cache on a hash of the whole argument buffer, making a padded encoding a cache miss rather than a wrong hit.

**Fix:** Out of scope for this branch — worth an issue. Pinning both lengths to their `SIZE` constants closes the length channel but not the legacy-slot one, so a complete fix also rejects non-zero bytes in the retired slots inside `VerifierKey::from_bytes`. A test that only pads the key segment passes while that hole is open; it has to mutate the retired slots too.

moCello (approved):
APPROVE — Both earlier points are addressed: the changelog entry and link are in place with the conflict markers gone, and the redundant count condition is removed while the decoder's early reject stays. The remaining validation matches exactly what the compiler produces, and the seven malformed-blob cases plus the round trip cover every reject branch and the positive path.

One note below, on encoding canonicality. Its fix lands in an issue, not on this branch, so nothing here asks for a change.

src/compiler/verifier.rs
HDauven: Tracked separately in #965, covering both fixed-length enforcement and non-zero retired-slot rejection.

PR #962, opened 2026-08-31
Derive BitIterator8 length from bytes

Neotamandua:
This PR is a very small change to the `BitIterator8` to derive its bit length from the actual byte slice length instead of `size_of::<E>()`. It also adds a small regression test covering the `Vec<u8>` case.

This makes it work correctly for dynamically sized byte containers such as `Vec<u8>`, while preserving existing behavior for fixed-size byte arrays.

### DIFF
--- a/src/bit_iterator.rs
+++ b/src/bit_iterator.rs
@@ -6,30 +6,21 @@
 
 //! Code taken from zcash repo and generalized as we do not have access to the
 //! limbs
-use core::mem;
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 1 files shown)


HDauven (approved):

moCello (approved):
APPROVE — The length derivation is now correct for every `AsRef<[u8]>` container, is bit-identical to the old arithmetic for the `[u8; 32]` inputs all existing callers use, and also removes a division-by-zero on empty input; the regression test fails on the pre-change code, and the crate-private scope means no changelog entry is needed.

PR #963, opened 2026-09-01
Avoid per-scalar serialization allocations

### LINKED ISSUE #964: Avoid per-scalar allocations during serialization
Avoid temporary per-element allocations when serializing polynomials, evaluations, and commitment keys.

The affected serialization paths currently allocate a temporary `Vec` for each scalar or G1 point before collecting the bytes. Instead, write each serialized element directly into a single preallocated output buffer.

This applies to:
- `Polynomial::to_var_bytes`
- `Evaluations::to_var_bytes`
- `CommitKey::to_var_bytes`

The serialized format and behavior should remain unchanged.

Neotamandua:
This PR avoids per-scalar allocations when serializing polynomials and evaluations. It replaces temporary `Vec` allocations for each scalar with a single preallocated output buffer and direct byte copies.

The serialized format and behavior remain unchanged.

Resolves #964

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Avoid per-element allocations when serializing polynomials, evaluations,
   and commitment keys [#964]
+- Bind proof construction to the public-input rows recorded at compile time [#957]
+- Reject serialized provers that predate the compiled public-input row layout [#957]
 - Update Criterion to 0.8 and align the dev-only `itertools` dependency [#953]
 - Raise the MSRV to Rust 1.96.1 [#951]
 - Update `blake2b_simd` to 1.0.5 [#950]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Avoid per-element allocations when serializing polynomials, evaluations,
+  and commitment keys [#964]
 - Update Criterion to 0.8 and align the dev-only `itertools` dependency [#953]
 - Raise the MSRV to Rust 1.96.1 [#951]
 - Update `blake2b_simd` to 1.0.5 [#950]
@@ -788,11 +790,9 @@ is necessary since `rkyv/validation` was required as a bound.
 - Proof system module.
 
 <!-- ISSUES -->
-<<<<<<< HEAD
+[#964]: https://github.com/dusk-network/plonk/issues/964
 [#959]: https://github.com/dusk-network/plonk/issues/959
-=======
 [#957]: https://github.com/dusk-network/plonk/issues/957
->>>>>>> 7fdedc29 (Bind proofs to compiled public-input rows)
 [#953]: https://github.com/dusk-network/plonk/issues/953
 [#951]: https://github.com/dusk-network/plonk/issues/951
 [#950]: https://github.com/dusk-network/plonk/issues/950
--- a/src/fft/evaluations.rs
+++ b/src/fft/evaluations.rs
@@ -50,12 +50,13 @@ pub(crate) struct Evaluations {
 impl Evaluations {
     /// Given an `Evaluations` struct, return it in it's byte representation.
     pub fn to_var_bytes(&self) -> Vec<u8> {
-        let mut bytes: Vec<u8> = self.domain.to_bytes().to_vec();
-        bytes.extend(
-            self.evals
-                .iter()
-                .flat_map(|scalar| scalar.to_bytes().to_vec()),
+        let mut bytes = Vec::with_capacity(
+            EvaluationDomain::SIZE + self.evals.len() * BlsScalar::SIZE,
         );
+        bytes.extend_from_slice(&self.domain.to_bytes());
+        for scalar in &self.evals {
+            bytes.extend_from_slice(&scalar.to_bytes());
+        }
 
         bytes
     }
--- a/src/fft/polynomial.rs
+++ b/src/fft/polynomial.rs
@@ -139,13 +139,12 @@ impl Polynomial {
     /// Given a [`Polynomial`], return it in it's bytes representation
     /// coefficient by coefficient.
     pub fn to_var_bytes(&self) -> Vec<u8> {
-        let degree = self.degree();
-        self.coeffs
-            .iter()
-            .enumerate()
-            .filter(|(i, _)| *i <= degree)
-            .flat_map(|(_, item)| item.to_bytes().to_vec())
-            .collect()
+        let len = self.coeffs.len().min(self.degree() + 1);
+        let mut bytes = Vec::with_capacity(len * BlsScalar::SIZE);
+        for scalar in &self.coeffs[..len] {
+            bytes.extend_from_slice(&scalar.to_bytes());
+        }
+        bytes
     }
 
     /// Generate a Polynomial from a slice of bytes.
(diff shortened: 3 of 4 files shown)


moCello (commented):
COMMENT — both rewrites are byte-for-byte equivalent to what they replace, including the polynomial encoder's trailing-zero truncation and its zero-polynomial case, and the length contract that `ProverKey::serialization_size` and `BufferWriter::finish` enforce holds unchanged.

Nothing gates the merge, and no finding disputes the code that landed.

One item is worth settling before merge: this change has no tracking issue and no `[Unreleased]` CHANGELOG entry, and neither can be added in place once it is merged.

One note accompanies it, small enough to fold into that same push: one instance of the per-element allocation this PR removes survives in code the PR did not touch.

src/fft/polynomial.rs
```diff
@@ -139,13 +139,12 @@ impl Polynomial {
...
-            .iter()
-            .enumerate()
-            .filter(|(i, _)| *i <= degree)
-            .flat_map(|(_, item)| item.to_bytes().to_vec())
-            .collect()
+        let len = self.coeffs.len().min(self.degree() + 1);
+        let mut bytes = Vec::with_capacity(len * BlsScalar::SIZE);
+        for scalar in &self.coeffs[..len] {
+            bytes.extend_from_slice(&scalar.to_bytes());
+        }
```
moCello, line 146: **NOTE:** One instance of the same per-element allocation remains; it is out of this PR's scope, so a follow-up issue settles it.

**Context:** `CommitKey::to_var_bytes` (`src/commitment_scheme/kzg10/key.rs:303-308`) still uses `flat_map(|item| item.to_bytes().to_vec()).collect()` over `powers_of_g` — the last per-element instance of the pattern in the crate, one allocation per SRS power. `CommitKey` is crate-private, but the path is reachable from the public `PublicParameters::to_var_bytes`. The sibling `CommitKey::to_raw_var_bytes` already has the shape this PR adopts, so the change is mechanical: `Vec::with_capacity(self.powers_of_g.len() * G1Affine::SIZE)` plus `extend_from_slice`.

**Fix:** Open an issue for it so the pattern does not survive as the only exception, or fold it into this PR alongside the changelog commit.

src/fft/evaluations.rs
```diff
@@ -50,12 +50,13 @@ pub(crate) struct Evaluations {
...
     /// Given an `Evaluations` struct, return it in it's byte representation.
     pub fn to_var_bytes(&self) -> Vec<u8> {
-        let mut bytes: Vec<u8> = self.domain.to_bytes().to_vec();
-        bytes.extend(
-            self.evals
-                .iter()
-                .flat_map(|scalar| scalar.to_bytes().to_vec()),
+        let mut bytes = Vec::with_capacity(
+            EvaluationDomain::SIZE + self.evals.len() * BlsScalar::SIZE,
         );
```
moCello, line 55: **CAUTION:** This change has no tracking issue, and no `[Unreleased]` CHANGELOG entry.

**Context:** CONTRIBUTING asks that PRs be linked to a previously opened issue; this one references none. Separately, the speedup is reachable from the public API through `Prover::to_bytes` and `Prover::serialized_size`, and comparable speed-only work is recorded under `### Changed` — [#919], [#920], [#921], [#925], [#927], [#933].

**Severity-rationale:** Raised pre-merge because a changelog entry added after the fact costs a separate PR.

**Fix:** Open an issue for the optimization and link this PR to it. Then add one `### Changed` line under `[Unreleased]` naming the change and referencing that issue, with the matching link definition at the bottom of the file.

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Avoid per-element allocations when serializing polynomials, evaluations,
+  and commitment keys [#963]
```
moCello: **CAUTION:** The changelog entry references this PR, not the tracking issue; the issue is #964.

**Context:** The link definition at `CHANGELOG.md:793` resolves to `.../issues/963`, which GitHub redirects to this pull request. Every other reference in the file points at an issue — eight checked, none a PR. CONTRIBUTING requires each PR to be linked to a previously opened issue, and #964 is that issue.

**Severity-rationale:** Worth correcting before merge because the entry is permanent once released, and changing it afterwards costs a PR of its own.

**Fix:** Change the entry's reference to `[#964]` and the link definition to `[#964]: https://github.com/dusk-network/plonk/issues/964`.

moCello (commented):
COMMENT — all three rewrites produce byte-identical output for every input, including the polynomial encoder's truncation at `degree()` and its empty-`coeffs` case, capacities are exact, the `ProverKey` size contract in `widget.rs` still holds, and both items from the previous round landed.

Nothing gates the merge. One comment below: the changelog entry links this PR (#963) rather than the tracking issue (#964), and that is cheaper to correct now than after release.

Two observations, neither needing a code change on this branch:

- The diff also removes a committed merge-conflict marker block from the changelog's link-definition section. The resolution is correct — both `[#959]` and `[#957]` are kept — but that block is still on `master` and stays there until this merges, so it is worth fixing on `master` independently rather than depending on this PR landing.
- Issue #964 and the PR description both describe polynomials and evaluations only, while the diff also rewrites `CommitKey::to_var_bytes`, which serializes G1 points rather than scalars. The changelog entry names all three sites, so the issue text is the one that has drifted; widening it keeps the record matching what merges.

CHANGELOG.md
moCello: **CAUTION:** The entry cites `[#963]` while the only definition added is `[#964]`, so the reference has no target and renders as literal text.

**Context:** `[#964]: https://github.com/dusk-network/plonk/issues/964` sits at `CHANGELOG.md:793`; no `[#963]:` definition exists in the file. That leaves a dangling reference in the entry and an unused definition at the bottom. Entries here link the tracking issue, not the PR.

**Severity-rationale:** The changelog is permanent once released, and correcting it afterwards costs a PR of its own — more than fixing it in the hunk that already touches this file.

**Fix:** Change `[#963]` at the end of the entry to `[#964]`.

moCello (commented):
COMMENT — the code is clean. All three rewrites are byte-identical to what they replace for every input, including the empty polynomial, capacities are exact, and the `ProverKey` length contract holds: every polynomial reaching `serialization_size` comes from `from_coefficients_vec` or `from_slice`, both of which normalize, so `to_var_bytes` never truncates below the counted length. Nothing in the code gates the merge.

One finding, on the changelog: the reference fix from last round landed on the link definition but not on the entry body, so `[#963]` now points at nothing while `[#964]` is defined and unused. Worth correcting here rather than in a follow-up PR, since this branch already owns the hunk.

The changelog hunk also removes the committed merge-conflict marker block from the link-definition section. The resolution is correct, and the block is still on `master`, where the same merge left `[#957]` defined with no entry referencing it — both worth fixing on `master` directly rather than widening this PR.

moCello:
The link-definition block this hunk touches also strands `[#957]`: `53f63c7` added the definition with no entry referencing it, so that change is undocumented and the definition renders unused.

The earlier note suggested leaving this to a separate PR against `master`. That assumed this branch wasn't already touching the block — it is, and it removes the markers around that definition, so the entry is two lines here instead.

Under `### Changed`:

```markdown
- Bind proof construction to the public-input rows recorded at compile time
  [#957]
- Reject serialized provers that predate the compiled public-input row layout
  [#957]
```

Two entries rather than one because the serialization break is a separate fact from the row binding: `Prover::to_var_bytes` gained a `public_input_indexes` section and `from_slice` rejects input without it, so provers serialized before `53f63c7` no longer load.

moCello (approved):
APPROVE — the three rewrites produce byte-identical output to what they replaced, including the empty and all-zero polynomial cases, with exact preallocation and no new imports or dependencies; the changelog entry cites the tracking issue with a matching definition and the earlier `[#957]` gap is closed. Nothing on this head asks for a change.

- The changelog's merge-conflict marker block is still on `master` until this merges; this branch's resolution of it is correct and needs no further change.

Issue #965, opened 2026-09-07
Enforce canonical verifier key encodings


HDauven:
Verifier deserialization currently accepts multiple byte encodings for the same key. The outer verifier and prover length fields are not pinned to the key `SIZE` constants, while the retired lookup commitment slots are serialized as zeroes but ignored during decoding.

Require the declared key lengths to match their fixed sizes and reject non-zero bytes in every retired slot. Add regressions covering both padded key segments and mutated retired slots.

This is canonical encoding hardening rather than a proof-soundness change. It also prevents equivalent verifier arguments from producing different raw-byte memoization keys.

Issue #942, opened 2026-08-20
Harden compressed-circuit bound invariants and parser coverage


HDauven:
## Summary

The compressed-circuit allocation fix in #939 uses correct bounds, but several invariants are mirrored manually and the new raw MessagePack header parser has uncovered rejection branches.

The packed-size constants restate the pinned msgpacker encoding and compressed-structure field counts. `Compiler::max_constraints` is also the hand-derived inverse of the compilation/trim capacity rule. Future changes could therefore make either bound too strict or too loose without a nearby test failing.

## Acceptance criteria

- Derive packed-size bounds from named encoding and structure components, including `SELECTORS_PER_POLYNOMIAL`.
- Use the selector-count constant for `CompressedPolynomial::scalar_indices` rather than repeating `11` in its return type.
- Pack a worst-case constraint and assert it fits the derived bound.
- Exercise `compile_with_compressed` at the exact capacity derived from real `PublicParameters`, and reject capacity plus one.
- Cover raw parser rejection for a non-array tag and truncated array header.
- Cover the `array32` (`0xdd`) header path with an in-bounds length.

Follow-up to #939 and [review discussion](https://github.com/dusk-network/plonk/pull/939#pullrequestreview-4982727840).

Issue #941, opened 2026-08-20
Semantically validate archived ProverKey values


HDauven:
## Summary

`ProverKey` derives rkyv `Archive` and structural `CheckBytes`, but the checked archive path does not apply the semantic validation performed by `ProverKey::from_slice`.

That means archived values can bypass the decoded degree bound, cached linear and vanishing coset-evaluation checks, and the serialized-length policy. This is analogous to the archived public-parameter validation addressed in #903 and should have an explicit semantic boundary of its own.

This is a pre-existing archive-path hardening gap identified during the review of #938; it does not block that focused fix.

## Acceptance criteria

- Identify every checked rkyv entry point that can construct or expose a `ProverKey`.
- Apply the same relevant semantic invariants enforced by checked byte deserialization before the value is accepted for proving.
- Reject malformed archived values without unwinding or allocating disproportionately.
- Add mutations for over-degree polynomials and inconsistent cached evaluations.
- Preserve valid archived representations, proving behavior, and `no_std` support.

Follow-up to #938 and [review discussion](https://github.com/dusk-network/plonk/pull/938#pullrequestreview-4982714751).

PR #968, opened 2026-09-09
Check serialized lengths before narrowing

### LINKED ISSUE #940: Normalize serialized-length conversions and truncation errors

HDauven:
## Summary

- Check serialized verifier sizes/indexes and raw commitment-key counts before narrowing to `usize`, matching the existing prover conversion policy.
- Report verifier byte-length multiplication/addition overflow as `BytesError(InvalidData)` instead of `NotEnoughBytes`, including on 64-bit targets. Preserve actual short-read errors and nested `BadLength { found, expected }` diagnostics.
- Keep fixed-length/retired-slot canonicality (#965), archived-key validation (#941), and compressed-circuit bounds (#942) out of this change.

Fixes #940.

## Validation

Rust 1.96.1, separate build directory for this branch:

- `make test`: 191 default + 207 all-feature release tests passed.
- Earlier i686-unknown-linux-musl release run with defaults + `rkyv-impl,legacy-proving`: 203 passed. The current documentation-only amendment passed the native suite; an extra i686 rerun was blocked by missing local musl C tooling.
- Both new 32-bit high-bit regressions fail against unchanged master and pass here.
- The extended public decoder regression covers multiplication, all three length additions, their representable short-read boundaries, and every truncated prefix of the frozen verifier. Separate 32-/64-bit negative controls reject the old multiplication policy and a multiplication-only fix that leaves addition overflow misclassified.
- `make fmt`, `make clippy`, `make no-std`, internal docs and committed diff checks passed; existing no-alloc dead-code warnings remain.
- Frozen upstream-Merlin V2/V3 fixtures pass. A local valid-proof probe confirms existing padded-key/retired-slot aliases remain accepted; no #965 policy change is included.

## Compatibility

The overflow error variant is an intentional public API behavior change, recorded in #940 and the changelog. Input acceptance is unchanged by this follow-up. Downstream callers and tests matching the previous overflow variant need adjustment, including Rusk's decoder-error regression. No Rusk source or dependency rollout is included, and no full Rusk replay was run.

No dependency, workflow, proof-version or serialization-writer changes.

### DIFF
--- a/src/commitment_scheme/kzg10/key.rs
+++ b/src/commitment_scheme/kzg10/key.rs
@@ -264,7 +264,8 @@ impl CommitKey {
 
         let mut len = [0u8; u64::SIZE];
         len.copy_from_slice(&bytes[..u64::SIZE]);
-        let len = u64::from_le_bytes(len) as usize;
+        let len = usize::try_from(u64::from_le_bytes(len))
+            .map_err(|_| dusk_bytes::Error::InvalidData)?;
 
         if len == 0 {
             return Err(dusk_bytes::Error::InvalidData.into());
--- a/src/error.rs
+++ b/src/error.rs
@@ -74,7 +74,9 @@ pub enum Error {
     PairingCheckFailure,
 
     // Serialization errors
-    /// Dusk-bytes serialization error
+    /// Dusk-bytes serialization error. Reads through dusk-bytes readers
+    /// retain `BadLength { found, expected }` diagnostics. Slice-length checks
+    /// in this crate return [`Error::NotEnoughBytes`] at any nesting depth.
     BytesError(DuskBytesError),
     /// This error occurs when there are not enough bytes to read out of a
     /// slice during deserialization.
--- a/tests/plonk_versioning.rs
+++ b/tests/plonk_versioning.rs
@@ -26,6 +26,60 @@ impl Circuit for MulCircuit {
 }
 
 #[test]
+fn verifier_truncation_preserves_top_level_error() {
+    let bytes = include_bytes!("fixtures/merlin-3/verifier.bin");
+    assert_eq!(Verifier::try_from_bytes(bytes).unwrap().to_bytes(), bytes);
+    for end in 0..bytes.len() {
+        assert!(matches!(
+            Verifier::try_from_bytes(&bytes[..end]),
+            Err(Error::NotEnoughBytes)
+        ));
+    }
+}
+
+#[cfg(target_pointer_width = "32")]
+#[test]
+fn verifier_rejects_truncated_high_bits() {
+    let bytes = include_bytes!("fixtures/merlin-3/verifier.bin");
+    Verifier::try_from_bytes(bytes).unwrap();
+    let word = |offset| {
+        u64::from_be_bytes(bytes[offset..offset + 8].try_into().unwrap())
+    };
+    // Six big-endian outer words, a little-endian inner key size, and
+    // a big-endian public-input index must all reject discarded high bits.
+    let key_start = 48 + word(0) as usize;
+    let index_start = key_start + word(8) as usize + word(16) as usize;
+    for offset in [3, 11, 19, 27, 35, 43, key_start + 4, index_start + 3] {
+        let mut mutated = bytes.to_vec();
+        mutated[offset] |= 1;
+        assert!(matches!(
+            Verifier::try_from_bytes(mutated),
+            Err(Error::BytesError(dusk_bytes::Error::InvalidData))
+        ));
+    }
+}
+
+#[cfg(target_pointer_width = "32")]
+#[test]
+fn prover_rejects_truncated_commit_key_length_high_bits() {
+    let mut rng = StdRng::seed_from_u64(63);
+    let pp = PublicParameters::setup(1 << 5, &mut rng).unwrap();
+    let (prover, _) = Compiler::compile::<MulCircuit>(&pp, b"length").unwrap();
+    let mut bytes = prover.to_bytes();
+    assert_eq!(Prover::try_from_bytes(&bytes).unwrap().to_bytes(), bytes);
+    let word = |offset| {
+        u64::from_be_bytes(bytes[offset..offset + 8].try_into().unwrap())
+            as usize
+    };
+    let commit_start = 56 + word(8) + word(16);
+    bytes[commit_start + 4] |= 1;
+    assert!(matches!(
+        Prover::try_from_bytes(bytes),
+        Err(Error::BytesError(dusk_bytes::Error::InvalidData))
+    ));
+}
+
+#[test]
 fn upstream_merlin_3_proofs_remain_valid() {
     let verifier = Verifier::try_from_bytes(include_bytes!(
         "fixtures/merlin-3/verifier.bin"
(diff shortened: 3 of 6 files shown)


src/commitment_scheme/kzg10/key.rs
```diff
@@ -264,7 +264,8 @@ impl CommitKey {
 
         let mut len = [0u8; u64::SIZE];
         len.copy_from_slice(&bytes[..u64::SIZE]);
-        let len = u64::from_le_bytes(len) as usize;
+        let len = usize::try_from(u64::from_le_bytes(len))
+            .map_err(|_| dusk_bytes::Error::InvalidData)?;
```
moCello, line 268: **NOTE:** `from_raw_var_bytes` now reports an unrepresentable count as `InvalidData` but still reports the multiplication and addition overflow of a representable count as `NotEnoughBytes`.

**Context:** The two `ok_or(Error::NotEnoughBytes)` calls sit a few lines below the new conversion, in the same function. A count that fits `usize` but exceeds `usize::MAX / G1Affine::RAW_SIZE` overflows the multiplication on both pointer widths. The path is reachable through `Prover::try_from_bytes`, which calls this decoder on the header-delimited commit-key segment. The verifier decoder in this PR was changed for the same reason. The polynomial reader in `ProverKey::from_slice` maps its coefficient-size multiplication overflow to `NotEnoughBytes` in the same way.

**Severity-rationale:** Pre-existing code, and the linked issue names other decoders as out of scope. Acceptance is unchanged, only the diagnostic is misleading.

**Fix:** Map both overflow branches in `from_raw_var_bytes` to `dusk_bytes::Error::InvalidData` and widen the second changelog entry to name the commit-key decoder. If that stays out of scope, record both remaining sites in a tracking issue.

src/error.rs
```diff
@@ -74,7 +74,9 @@ pub enum Error {
     PairingCheckFailure,
 
     // Serialization errors
-    /// Dusk-bytes serialization error
+    /// Dusk-bytes serialization error. Nested short reads retain
+    /// `BadLength { found, expected }` diagnostics rather than being collapsed
+    /// into [`Error::NotEnoughBytes`].
```
moCello: **NOTE:** The doc states a rule that the decoders do not follow for payload short reads.

**Context:** Only fixed-size reads through the dusk-bytes reader return `BadLength`. Length checks done by this crate return `NotEnoughBytes` for nested payloads too: `ProverKey::from_slice` returns `NotEnoughBytes` when a polynomial or evaluation payload is short, and the PR's own test in `widget.rs` asserts that outcome for `truncated_coefficients`. The linked issue asks for this decision to be documented, so the sentence is the record of that decision.

**Fix:** Reword to the actual split: reads through dusk-bytes readers keep `BadLength { found, expected }`; slice-length checks done by this crate return `NotEnoughBytes`, at any nesting depth.

tests/plonk_versioning.rs
```diff
@@ -25,6 +25,88 @@ impl Circuit for MulCircuit {
...
+            Verifier::try_from_bytes(bytes).err(),
+            Some(expected),
+            "header: {header:?}"
+        );
+    }
+}
+
+#[cfg(target_pointer_width = "32")]
+#[test]
+fn verifier_rejects_truncated_high_bits() {
```
moCello, line 70: **NOTE:** The two 32-bit regressions never compile or run in CI, because no workflow job tests a 32-bit target.

**Context:** `dusk_ci.yml` only runs `cargo check` for `thumbv6m-none-eabi`, which does not build tests. The test job and the coverage job run on the host. The same gap applies to `prover_rejects_truncated_commit_key_length_high_bits`. The PR description records a local i686 run, and a later change can break the checked conversions without a CI signal.

**Fix:** Add a job that runs the release tests on a 32-bit target, for example `cross test --release --target i686-unknown-linux-musl`, or open an issue to track that job.

moCello (approved):
APPROVE — every remaining truncating cast in the verifier, verifier-key and raw commit-key decoders is now a checked conversion, the overflow reclassification is recorded in the issue and the changelog, and the boundary tests cover multiplication, all three additions and their representable neighbours on both pointer widths. Three notes below, none of them gating and all small.

src/commitment_scheme/kzg10/key.rs
HDauven: Tracked in #971 so the remaining decoder error changes stay separate from this PR.

src/error.rs
HDauven: Clarified the split between dusk-bytes reader errors and PLONK's slice-length checks.

tests/plonk_versioning.rs
HDauven: Tracked in #970 for execution on a 32-bit CI target.

PR #969, opened 2026-09-09
Derive and test compressed-circuit bounds

### LINKED ISSUE #942: Harden compressed-circuit bound invariants and parser coverage

HDauven:
## Summary

- Derive the existing packed-size limits from named encoding/structure components and reuse the selector count. The limits remain 30 fixed bytes plus 857 per constraint.
- Test worst-case packing, malformed raw array headers and the in-bounds array32 path.
- Test exact compilation capacity and capacity-plus-one rejection using real public parameters, including non-power-of-two budgets.

Fixes #942.

Independently based on master, not stacked on #968.

## Validation

Rust 1.96.1, separate build directory for this branch:

- `make test` after rebasing onto merged #968: 194 default + 210 all-feature release tests passed.
- Earlier i686-unknown-linux-musl release run with defaults + `rkyv-impl,legacy-proving`: 203 passed. The added header regression passed on 64-bit; a fresh i686 build was blocked by missing local musl C tooling.
- Negative controls fail as intended: reducing the packed bound by one, reducing the array32 header allowance, reducing the fixed overhead and removing capacity floor rounding.
- `make fmt`, `make clippy`, `make no-std`, internal docs and committed diff checks passed; existing no-alloc dead-code warnings remain.
- Existing direct/compressed serialization parity and proof verification tests pass.

No runtime parser/compilation algorithm, public API, wire-format, dependency or workflow changes.

### DIFF
--- a/src/composer/compress.rs
+++ b/src/composer/compress.rs
@@ -43,7 +43,9 @@ pub struct CompressedPolynomial {
 }
 
 impl CompressedPolynomial {
-    fn scalar_indices(&self) -> [usize; 11] {
+    fn scalar_indices(
+        &self,
+    ) -> [usize; CompressedCircuit::SELECTORS_PER_POLYNOMIAL] {
         [
             self.q_m,
             self.q_l,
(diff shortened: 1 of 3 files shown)


src/composer/compress.rs
```diff
@@ -589,6 +598,67 @@ mod tests {
...
+                CompressedCircuit::SELECTORS_PER_POLYNOMIAL
+            ],
+            polynomials: vec![polynomial],
+            constraints: vec![constraint],
+        };
+        packed.clear();
+        worst.pack(&mut packed);
+        assert!(
+            packed.len() <= CompressedCircuit::packed_size_limit(1).unwrap()
+        );
```
moCello, line 646: **NOTE:** The whole-circuit assertion leaves 16 bytes of slack, so the fixed-bytes component of the bound is not pinned.

**Context:** The four vectors in `worst` hold one, eleven, one and one element, all inside the fifteen a fixarray encodes, so each header packs in one byte. On 64-bit targets the packed length lands 16 bytes under `packed_size_limit(1)`, which is exactly the slack the `array32` header allowance adds. Lowering `PACKED_VECTOR_HEADER_BYTES` or `PACKED_FIXED_BYTES` by up to 16 leaves this assertion green. Only the per-constraint assertion above it is tight.

**Fix:** Add an exact check under `#[cfg(target_pointer_width = "64")]`: `assert_eq!(packed.len() + 4 * (CompressedCircuit::PACKED_VECTOR_HEADER_BYTES - 1), CompressedCircuit::packed_size_limit(1).unwrap())`. The gate is needed because `usize::MAX` packs in five bytes on 32-bit targets, which the PR validates. Alternatively drop the second assertion so the test claims no more than it checks.

moCello (approved):
APPROVE — The derived constants match the pinned msgpacker encoding byte for byte, the numeric limits are unchanged, and the capacity test pins the floor rounding and the capacity-plus-one rejection against real public parameters. One note below, small and yours to take or leave.

src/composer/compress.rs
HDauven: Added the exact 64-bit bound check and independently pinned the actual array32 header. Reducing either allowance now fails the test.

PR #972, opened 2026-09-11
Derive structural rkyv validation for proofs and verifier keys

### LINKED ISSUE #973: Remove handwritten unsafe code from rkyv validation
Replace handwritten unsafe rkyv & bytecheck validation code with derive-based and checked APIs where possible.

The goal is to reduce the amount of unsafe code in the `rkyv-impl` serialization/deserialization paths without changing any existing validation semantics, archive formats, public APIs, or runtime behavior.

Custom validation can remain when semantic checks cannot be expressed by the derive macros alone.

Neotamandua:
This PR replaces the handwritten `unsafe` rkyv validation code with derive-based `CheckBytes` and safe checked archive access.

By that we remove the custom field-by-field validation helpers from `Proof` and `VerifierKey` and also preserve the existing custom validators where semantic checks are still required.

This PR leaves public API, serialization format, or runtime behavior unchanged.

### DIFF
--- a/src/commitment_scheme/kzg10/key.rs
+++ b/src/commitment_scheme/kzg10/key.rs
@@ -105,6 +105,8 @@ fn archived_commit_key_points_are_valid(
     powers.iter().all(archived_g1_is_valid)
 }
 
+// bytecheck 0.6 derives only structural checks. Keep this custom validator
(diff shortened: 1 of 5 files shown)


HDauven (approved):
LGTM

PR #976, opened 2026-09-21
Fix wires blinding

### LINKED ISSUE #975: Fix wire blinding
#### Summary
The blinding of the wires a b d should be increased by 1. As a reference, we got this fixed already in: https://github.com/dusk-network/zk-tools/pull/14/changes/7c75acde36d202efac22fa49f1d981c5576d9089

xevisalle:
Closes: #975


HDauven (approved):
LGTM

src/commitment_scheme/kzg10/srs.rs
```diff
@@ -47,11 +47,11 @@ pub struct PublicParameters {
...
-    /// The maximum degree is the degree of the constraint system + 6,
-    /// because adding the blinding factors requires some extra elements
-    /// for the SRS: +1 per each wire (we have 4 wires), plus +2 for the
-    /// permutation polynomial
-    pub(crate) const ADDED_BLINDING_DEGREE: usize = 6;
+    /// The permutation numerator has degree at most 5n + 9: the masks
+    /// contribute 2 each for a, b, d and z, and 1 for c. Division by X^n - 1
+    /// leaves degree 4n + 9. Splitting the quotient at n, 2n and 3n therefore
+    /// requires SRS powers through n + 9 for the final chunk.
+    pub(crate) const ADDED_BLINDING_DEGREE: usize = 9;
```
moCello, line 54: **CAUTION:** Public parameters from earlier releases now compile only circuits up to half their domain, and the pinned production CRS stops covering three of the four transfer circuits.

**Context:** `trim` asks the commit key for `truncated_degree + 9`, and `Compiler::max_constraints` subtracts 9 from the available degree before it floors the result to a power of two. That floor is what turns three degrees into an octave. A serialized `setup(2^k)` from an earlier release holds `2^k + 6`, so the available degree becomes `2^k - 3`, which floors to `2^(k-1)`. The test `sufficiently_large_legacy_setup_and_prover_key_remain_usable` shows the same step at small scale: a `32 + 6` setup serves a 16-row domain.

**Impact:** The CRS that `rusk-profile` pins through `CRS_17_HASH` holds 131,079 powers, so its maximum degree is `2^17 + 6 = 131,078`. Against it, `Compiler::max_constraints` returns 131,066 before this change and 65,536 after, because `131,078 - 9 = 131,069` lands three short of `2^17`. Three of the four cached transfer circuits sit above the new limit, at 73,646, 100,971 and 128,296 constraints, and each stops at `TruncatedDegreeTooLarge`. Only the 46,321-constraint circuit still compiles.

Removing the row padding does not offset this. That saving applies before the rounding, so it reaches only a circuit within six gates of a power of two. The blinding allowance applies after it, at every circuit size. A ceremony CRS cannot get more powers either, so the entry's remedy, to regenerate from sufficiently large public parameters, does not apply to it.

The release number decides which way it fails. `rusk-profile` derives key ids from the circuit and the plonk major and minor version, not the patch version. A minor bump changes the ids, so key generation recompiles and stops at `TruncatedDegreeTooLarge`. A patch release keeps the ids, so a profile directory that already holds the keys skips the recompile. Each cached key carries its own commit key at maximum degree `2^17 + 6`, and a proof with one then fails with `PolynomialDegreeTooLarge` at proving time.

**Severity-rationale:** The break is certain rather than conditional, because the CRS is hash-pinned and the circuit domains are recorded in the cached keys. It does not gate: downstream key generation compiles all four circuits against the CRS, so the next version bump stops there rather than in production. Recovering the three degrees inside the crate would mean moving the quotient split off n, 2n and 3n, which changes the equation the verifier recomputes, and this change deliberately keeps proofs and verifier keys compatible.

**Fix:** Release under a minor bump, so the derived key ids change and the keys rebuild at key generation instead of failing at proving time. Open an issue for the three-power shortfall against the pinned CRS, and link it from the changelog entry. Smaller circuit domains do not close it: the next domain down is `2^16`, which needs the two larger circuits to shed more than a third of their gates.

moCello (commented):
COMMENT — The blinding fix is correct. Wires a, b and d get three random terms, c gets two, and the quotient bound of 4n + 9 holds for every gate. Verifier keys and proof serialization stay compatible. Nothing gates the merge. The capacity break below costs a coordination step downstream, and the changelog states a remedy that a ceremony CRS cannot follow.

- The issue asks only to raise the a, b and d blinding by one. The change also removes the six-row padding from compilation, which changes prover key sizes and compile capacity on its own.

CHANGELOG.md
```diff
@@ -100,6 +100,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix shifted-wire blinding and unnecessary SRS oversizing, adapted from
+  [zk-tools#14](https://github.com/dusk-network/zk-tools/pull/14/commits/7c75acde36d202efac22fa49f1d981c5576d9089).
+  Proving requires key capacity through degree `n + 9` (previously `n + 6`),
+  where `n` is the evaluation-domain size. Undersized proving keys must be
+  regenerated from sufficiently large public parameters; proof serialization
+  and verifier keys remain compatible, while deterministic proof bytes change.
+  [#975]
```
moCello: **CAUTION:** One entry holds four facts, a remedy and an inline link to another repository; the capacity change also belongs under Changed, not Fixed.

**Context:** The entry mixes the blinding fix, the removed row padding, the new `n + 9` requirement and a compatibility statement. Every other entry cites its issue or pull request through a `[#nnn]` reference resolved at the end of the file. This entry carries the only inline link in any entry body, and it points at a commit inside another repository's pull request. The remedy and the compatibility notes belong in #975.

**Fix:** Keep one sentence under `### Fixed` for the wire blinding with `[#975]`. Move the capacity change to `### Changed`, and state there that a `2^k` setup now tops out at a `2^(k-1)` domain. Remove the inline zk-tools link.

src/commitment_scheme/kzg10/srs.rs
xevisalle: Solved in: https://github.com/dusk-network/trusted-setup/issues/20

